PR #168, opened 2026-03-01
Exp 6: Data volume learning curve + injury timing diagnostic

Acusick1:
## Summary

- Add `experiments/scripts/exp6_learning_curve.py` — standalone experiment script
- **Learning curve**: R² plateaus at ~1,500 events (oscillates 0–0.035 from 1.5K→4.5K). Log-fit dR²/dN = 6.7e-06 at max N. More OddsPortal data will not meaningfully improve the model.
- **Injury timing diagnostic (2×2)**: injuries add nothing at sharp (12-24h) or pregame (3-12h) tier. 0-3h closing window untestable with OddsPortal data.
- Update `docs/MODELING.md` — What We Know, Experiment Plan (Exp 6 → completed), Experiment Log, Open Questions

## Key numbers

| Subset | R² Mean | R² Std |
|--------|---------|--------|
| 500 events | -0.066 | 0.096 |
| 1,500 events | +0.036 | 0.079 |
| 3,000 events | +0.014 | 0.046 |
| 4,524 events | +0.018 | 0.055 |

| Feature Group | Tier | R² Mean |
|---------------|------|---------|
| tabular | sharp | +0.018 |
| tabular | pregame | +0.001 |
| tabular+injuries | sharp | +0.006 |
| tabular+injuries | pregame | +0.002 |

## Test plan

- [x] `uv run python experiments/scripts/exp6_learning_curve.py` runs to completion
- [x] `learning_curve.csv` has 10 rows; `injury_timing.csv` has 4 rows
- [x] Plots render correctly (gitignored, generated locally)
- [x] Ruff check + format passes

### DIFF
--- a/docs/MODELING.md
+++ b/docs/MODELING.md
@@ -213,7 +213,7 @@ Time-series per snapshot:
 
 ### Signal
 - ~3.6% R² is the ceiling for public sportsbook features at 5K events, with the learning curve plateaued. Can Polymarket cross-source features push it higher?
-- ~~Injury features dominate importance~~ — **Answered**: injuries add zero signal when properly tuned. Earlier apparent importance was a tuning artifact.
+- ~~Injury features dominate importance~~ — **Partially answered**: injuries add zero signal at ≥3h decision times when properly tuned (earlier apparent importance was a tuning artifact). GTD injury at 0-3h (r=0.28, Exp 4) remains untested — OddsPortal lacks closing-window snapshots.
 - Does the signal generalize to other sports, or is it NBA-specific?
 - Do PM features add signal? Untested at scale — only tested with 230 events (insufficient data). PM order flow from CLOB snapshots remains untapped.
 - Does sharp-retail divergence (Pinnacle vs DraftKings/FanDuel) contain more signal than cross-source (PM vs SB)?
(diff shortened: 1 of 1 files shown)


Acusick1 (commented):
## Review: Exp 6 Learning Curve + Injury Timing

Overall this is clean work that follows established patterns well. A few issues to address:

### Bugs

**1. `plot_learning_curve` return type mismatch**
Annotated as `-> None` but returns `popt`. Should be `-> tuple[float, float] | None`.

**2. `plot_injury_timing` color logic is no-op**
```python
colors = ["#2196F3" if "tabular" in label else "#FF9800" for label in labels]
```
All 4 labels contain "tabular" (both `"tabular\nsharp"` and `"tabular+injuries\nsharp"`), so every bar is blue. Needs something like `"tabular+" not in label` or `label.startswith("tabular\n")` to distinguish the groups.

**3. Fragile `hours_until_event` column index**
```python
hours = X_sub[:, -1]  # hours_until_event is the last feature column
```
Silent wrong data if feature order ever changes. Should assert `feature_names[-1] == "hours_until_event"`.

### Design concern: pregame diagnostic doesn't cleanly test the timing hypothesis

`TierSampler(decision_tier="pregame")` is a *floor*, not a *band* — it accepts pregame, sharp, early, and opening tiers, then picks the most recent by wall-clock time. For OddsPortal events with only opening (~19h/sharp) and closing (~0.1h/closing) snapshots:
- The closing snapshot is excluded (closer than pregame tier)
- The opening snapshot at sharp tier is the only candidate — **same as `decision_tier=sharp`**

Only the 335 events that actually have a pregame-tier snapshot (~7% of 4,857) sample at 3-12h. The other 93% sample identically to the sharp experiment. The R² difference (0.001 vs 0.018) could be those 335 events dragging down the average, not a clean timing effect.

This is worth documenting in FINDINGS.md — the pregame diagnostic is diluted, not a controlled comparison of decision times. A note like "~93% of pregame-tier events fall back to sharp-tier snapshots (~19h); only 335 events sample at actual pregame timing" would prevent over-interpreting the results.

### Minor

- CI is still pending — wait for green before merge.

PR #171, opened 2026-03-01
Add Exp 7b: line shopping across bookmakers

Acusick1:
## Summary

- Re-run Exp 7 walk-forward betting simulation selecting best available price across 4 UK bookmakers (bet365, betway, betfred, bwin) instead of bet365-only
- Effective vig drops from 4.13% to 1.59% (2.55pp reduction), ROI improves +0.3–1.3pp at every threshold, but still not profitable at thresholds ≤0.03
- bet365 provides best price 67% of the time, limiting shopping upside — the ~2.5pp vig reduction is smaller than the ~4% gap between signal strength and breakeven

## Key Results

| Threshold | bet365 ROI | Best Available ROI | Improvement |
|-----------|-----------|-------------------|-------------|
| 0.005 | -4.21% | -2.88% | +1.34pp |
| 0.010 | -2.42% | -1.49% | +0.93pp |
| 0.015 | -6.01% | -5.51% | +0.50pp |
| 0.020 | -9.31% | -8.79% | +0.52pp |
| 0.030 | -11.75% | -11.28% | +0.47pp |
| 0.050 | +3.36% | +3.68% | +0.32pp |

## Test plan

- [x] Script runs to completion against local DB
- [x] `threshold_sweep.csv` has 12 rows (6 thresholds × 2 modes)
- [x] bet365-only results match Exp 7 exactly
- [x] Best-available ROI ≥ bet365-only at every threshold
- [x] `ruff check --fix && ruff format` passes


Acusick1 (commented):
## Review: Exp 7b Line Shopping Across Bookmakers

**CI**: Passing

### General
- [x] Correct methodology — same predictions, same bets, different price selection. Clean A/B comparison.
- [x] Results verified: bet365-only numbers in threshold_sweep.csv match Exp 7 exactly
- [x] Best-available ROI >= bet365-only at every threshold (monotonic improvement, as expected)

### Code Quality
- [x] Clean `BettingContext` redesign — `list[BookmakerOdds]` per side instead of single int
- [x] `_pick_odds()` helper with `line_shop` flag is a nice abstraction — single dispatch point for the mode toggle
- [x] `compute_vig_stats()` correctly computes best-available overround by mixing best home from any book with best away from any book (cross-book shopping)
- [x] Bookmaker frequency analysis adds useful insight (bet365 = 67% best price)
- [x] `simulate_betting` runs once per (mode, threshold) — the duplicate call pattern from Exp 7 (sweep + all_bets) persists here but is minor

### Type Safety & Patterns
- [x] Complete type hints, modern syntax
- [x] Proper async/sync separation
- [x] `strict=True` on `zip(bars, df.iterrows())` — good practice

### Observations (non-blocking)

1. **Permutation test p-values are *worse* with line shopping** (e.g., threshold=0.05: p=0.260 → p=0.334). This makes sense — shuffled predictions also benefit from line shopping, so the null distribution shifts right, and the observed ROI doesn't separate further. Worth noting in interpretation: line shopping helps absolute ROI but doesn't improve statistical significance because the null also improves.

2. **`compute_vig_stats` uses `calculate_implied_probability` for bet365 but `1.0 / decimal_odds` for best-available** (lines 428-437). Both are correct (they should produce the same result), just an inconsistency in style. `calculate_implied_probability` internally does `1/decimal_odds` anyway.

3. **`simulate_betting` called 3 extra times** beyond the sweep loop: once for `all_bets.csv`, once for bookmaker frequency, once for equity curves. These are all at different thresholds/modes so can't easily be consolidated, just noting it's ~O(n) redundant work on 1.3K bets (negligible).

### Result interpretation

The headline finding is clean: 2.55pp vig reduction, +0.3-1.3pp ROI improvement, still not profitable. bet365 being the best price 67% of the time quantifies why shopping within 4 UK books has limited upside. The conclusion that you'd need exchange-level costs or stronger signal is well-supported.

LGTM — well-structured experiment with a clear negative result that narrows the path forward.

PR #204, opened 2026-03-08
Parameterize Lambda jobs by league via EventBridge payload

### LINKED ISSUE #202: Parameterize Lambda jobs by league via EventBridge payload
## Goal

Adding a new league (La Liga, Bundesliga) should require only infrastructure changes (EventBridge rules, model in S3) — no code changes. The three league-specific jobs (`score-predictions`, `daily-digest`, `fetch-oddsportal-results`) currently hardcode `SPORT_KEY = "soccer_epl"` and need to accept league parameters from the event payload.

## Location

- `packages/odds-lambda/odds_lambda/jobs/score_predictions.py`
- `packages/odds-lambda/odds_lambda/jobs/daily_digest.py`
- `packages/odds-lambda/odds_lambda/jobs/fetch_oddsportal_results.py`
- `deployment/aws/terraform/eventbridge.tf`

## Scope

- Add `sport_key` and `model_name` as optional kwargs to `score_predictions.main()`, falling back to env vars / defaults when absent
- Add `sport_key` and `model_name` as optional kwargs to `daily_digest.main()`, derive embed title from sport_key
- Add `sport_key` (or league spec identifier) as optional kwarg to `fetch_oddsportal_results.main()`, replacing the module constants
- Update EventBridge rules for `score-predictions` and `daily-digest` to pass `sport_key` and `model_name` in the payload
- Add `fetch-oddsportal-results` EventBridge payload to include `sport_key`

## Success criteria

- [ ] All three jobs accept `sport_key`/`model_name` from event payload and use them for DB queries and model loading
- [ ] Existing behavior preserved when kwargs are omitted (EPL defaults)
- [ ] Adding a new league requires only: new `LeagueSpec` entry, trained model in S3, new EventBridge rules with league-specific payloads
- [ ] No new Lambda functions or Docker images needed per league
- [ ] Digest embed title reflects the league (e.g., "La Liga Daily Digest" not "EPL Daily Digest")

## Notes

- `fetch_oddsportal.py` already iterates `LEAGUE_SPECS` and is multi-league ready — no changes needed there
- The lambda handler already passes extra event payload fields as kwargs to jobs that declare `**kwargs`
- Depends on #201 for fully zero-code inference, but job parameterization can land independently

Acusick1:
## Summary
- All three league-specific jobs (`score-predictions`, `daily-digest`, `fetch-oddsportal-results`) now accept `sport_key` and/or `model_name` from EventBridge event payloads, falling back to EPL defaults when omitted
- Lambda handler updated to forward kwargs matching declared parameter names (not just `**kwargs`), so jobs can use explicit typed parameters
- EventBridge rules updated with `sport_key`/`model_name` in payloads; adding a new league requires only new rules + model in S3
- Digest embed title derived from `sport_key` via `LEAGUE_DISPLAY_NAMES` mapping (e.g. "La Liga Daily Digest")

## Test plan
- [x] All 51 existing unit tests pass
- [ ] Deploy and verify EPL jobs still work with explicit payload params
- [ ] Verify adding a second league (e.g. La Liga) works by adding a `LeagueSpec` entry + EventBridge rules

Closes #202


Acusick1 (commented):
## Review — PR #204: Parameterize Lambda jobs by league via EventBridge payload

**CI:** ✅ Passing

### Overall

Clean PR — the parameterization approach is sound and the handler change is a strict improvement. A few items to flag:

### Issues

1. **`LEAGUE_DISPLAY_NAMES` duplicates `LeagueSpec.sport_title`** (`daily_digest.py:32-38`)
   `LeagueSpec` already has `sport_title` (e.g. `"EPL"`). `LEAGUE_DISPLAY_NAMES` is a second source of truth for the same mapping. Consider importing `LEAGUE_SPECS` and deriving the display name from there (like `fetch_oddsportal_results.py` already does with `_get_league_spec()`), or at minimum document that both must be kept in sync.

2. **Handler docstring is stale** (`lambda_handler.py:37-38`)
   > "Jobs that accept parameters should declare **kwargs to receive them; jobs that don't will ignore extra fields (we inspect the signature)."
   
   This is now backwards — jobs should declare *explicit* parameters, not `**kwargs`. The handler filters by declared param names. The docstring should reflect the new contract.

3. **`score-predictions` EventBridge rule is missing `sport_key`** (`eventbridge.tf:87`)
   `score_predictions.score_events()` queries events by sport — it uses `MODEL_NAME` env var to load the model, but the underlying query filters on sport. Currently it's inherited from `score_events()`'s internal logic (via config or hardcoded). Is there an implicit EPL filter in `score_events()` you're relying on, or should `sport_key` be passed here too for when a second league is added?

### Minor

- `run_harvester_historic()` default `spec=None` + immediate lookup is fine but slightly unusual — a sentinel default of `DEFAULT_SPORT_KEY` with lookup at the call site would be more explicit. Not blocking.
- `accepted = {k: v for k, v in kwargs.items() if k in params}` — ternary on line 55 (`await job_fn(**accepted) if accepted else await job_fn()`) is equivalent to just `await job_fn(**accepted)` since `**{}` is a no-op. Simpler to always pass it.

### Checklist

- [x] CI passes
- [x] No compatibility hacks
- [x] Proper abstraction — reuses `LeagueSpec` from `fetch_oddsportal`
- [x] Type hints present
- [x] Modern syntax (`str | None`, `list[T]`)
- [x] No security issues
- [x] No hardcoded secrets
- [x] Timezone handling correct (UTC throughout)
- [ ] Stale docstring in handler (item 2)
- [ ] Dual source of truth for league display names (item 1)
- [ ] Verify `score-predictions` doesn't need `sport_key` (item 3)

Acusick1 (commented):
## Re-review

All three items from the previous review are addressed:

1. ✅ `LEAGUE_DISPLAY_NAMES` replaced with `_league_display_name()` deriving from `LeagueSpec.sport_title` (single source of truth, clean fallback)
2. ✅ Handler docstring updated to reflect the new param-matching contract
3. ✅ `run_harvester_historic()` now requires `LeagueSpec` explicitly (no sentinel default)
4. ✅ Ternary simplified — always passes `**accepted`

No remaining issues. LGTM pending CI ✅

Issue #270, opened 2026-04-09
Add paper trades table and settlement logic


Acusick1:
## Goal

The betting agent (docs/BETTING_AGENT.md, Phase 1) needs a persistent record of forward-looking paper bets — placed before events start, then settled against actual results. This bridges the existing backtest engine (historical) with live agent evaluation. The schema should align with `BetRecord` so backtest analysis tools (risk metrics, equity curve) can operate on paper trade history.

## Scope

- Add `PaperTrade` SQLModel in `odds_core/` with fields aligned to `BetRecord`: event_id, market, selection, bookmaker, odds, stake, reasoning, confidence, placed_at, settled_at, result, pnl, bankroll_before, bankroll_after
- Add alembic migration
- Add settlement service: query unsettled trades, match against final scores in events table, compute P&L
- Add portfolio queries: open bets, settled P&L summary, current bankroll, exposure by event/market
- Add CLI commands under `odds paper` (or similar): `list`, `settle`, `summary`

## Location

- `packages/odds-core/odds_core/` — new model file or extend `prediction_models.py`
- `packages/odds-lambda/odds_lambda/` or `packages/odds-cli/odds_cli/` — settlement service and CLI
- `packages/odds-analytics/odds_analytics/backtesting/models.py` — reference for `BetRecord` field alignment

## Implementation approach

- `PaperTrade` is a SQLModel table, not a dataclass — it persists across sessions
- Settlement logic should be idempotent (re-running settle on already-settled trades is a no-op)
- 3-way market support (home/draw/away) from day one — this is EPL
- Bankroll tracking: initial bankroll configurable, updated on settlement. Stake validation against available bankroll at placement time
- Align field names with `BetRecord` where possible so a conversion function can produce `BetRecord` instances for backtest analysis tools

## Success criteria

- [ ] `PaperTrade` table created with migration
- [ ] Paper bets can be inserted with reasoning and confidence
- [ ] Settlement resolves bets against actual scores (win/loss/push for 3-way)
- [ ] Portfolio queries return open bets, P&L, bankroll, exposure
- [ ] CLI commands for listing, settling, and summarising paper trades
- [ ] `PaperTrade` records can be converted to `BetRecord` for backtest metric calculation
- [ ] Settlement is idempotent

## Notes

- Existing paper trading PRs (#188/#190/#189) built the `Prediction` table and scoring pipeline — that's CLV signal logging, not bet tracking. This is the complementary piece.
[...]

PR #273, opened 2026-04-10
Build odds-mcp FastMCP server for betting agent

### LINKED ISSUE #271: Build odds-mcp FastMCP server for betting agent
## Goal

Phase 1 of the betting agent (docs/BETTING_AGENT.md). Expose existing jobs, DB queries, and paper trading as MCP tools so any LLM runtime (Claude Code, Pydantic AI, LangGraph) can interact with the system programmatically. This is the canonical tool interface for the agent.

## Scope

- Create a FastMCP server as a new package or module
- Implement MCP tools wrapping existing functionality:
  - [ ] `get_upcoming_fixtures(league, days_ahead)` — query events table
  - [ ] `get_current_odds(event_id)` — latest snapshots per bookmaker
  - [ ] `get_odds_history(event_id)` — odds movement timeline for an event
  - [ ] `refresh_scrape(league, market)` — trigger OddsPortal scrape on-demand
  - [ ] `get_model_prediction(event_id)` — run CLV model inference
  - [ ] `get_event_features(event_id)` — feature vector (standings, schedule, stats)
  - [ ] `paper_bet(event_id, market, selection, odds, stake, reasoning)` — record a paper trade
  - [ ] `get_portfolio()` — open bets, P&L summary, current bankroll
  - [ ] `settle_bets()` — settle completed events against actual results
- Add MCP server config to `.mcp.json` for Claude Code discovery
- Tools must return structured dicts/JSON, not log messages

## Dependencies

- #270 (paper trades table) for `paper_bet`, `get_portfolio`, `settle_bets`

## Implementation approach

- Use FastMCP (PyPI: `fastmcp`) — decorator-based tool registration
- Thin wrappers: MCP tools call into existing service/job code, not re-implement logic
- All tools are async, using `async_session_maker()` for DB access
- Server runs as a standalone process (stdio transport for Claude Code, HTTP for future runtimes)
- Return types should be serialisable dicts that both an LLM and a programmatic agent can parse

## Constraints

- Tools must be stateless — all state in DB or passed as arguments
- No session/runtime-specific state
- Follow existing async patterns (`async with async_session_maker() as session`)

## Success criteria

- [ ] FastMCP server starts and registers all tools
- [ ] Claude Code can discover and call tools via `.mcp.json`
- [ ] Each tool returns structured data an LLM can reason over
- [ ] `refresh_scrape` triggers actual OddsPortal job
- [ ] `get_model_prediction` loads model and runs inference
- [ ] Paper trading tools integrate with #270

## Notes

- This server is designed to be consumed by Claude Code now and Pydantic AI / LangGraph later — keep the MCP interface clean as it becomes the contract between agent runtime and data layer.
- `get_model_prediction` and `get_event_features` depend on model being published to S3 — should return a clear error if no model is available.

Acusick1:
## Summary

- New `packages/odds-mcp/` workspace package with FastMCP server exposing 9 MCP tools as the canonical agent interface
- **Fixture/odds tools**: `get_upcoming_fixtures`, `get_current_odds`, `get_odds_history` — thin wrappers over `OddsReader`
- **Scraping**: `refresh_scrape` — triggers OddsPortal scrape on-demand via `ingest_league()`
- **ML tools**: `get_model_prediction` (queries Prediction table), `get_event_features` (runs TabularFeatureExtractor with configurable sharp/retail bookmakers)
- **Paper trading**: `paper_bet`, `get_portfolio`, `settle_bets` — wraps paper trading module from #270
- `.mcp.json` for Claude Code stdio transport discovery
- 18 unit tests covering serialization helpers and error paths

## Design decisions

- `get_odds_history` uses `extract_odds_from_snapshot()` for structured output instead of raw JSON blobs (context-window friendly)
- `get_event_features` accepts `sharp_bookmakers`/`retail_bookmakers` params, defaulting to hybrid sharp reference `[pinnacle, betfair_exchange]`
- `refresh_scrape` uses `LEAGUE_SPECS` from `fetch_oddsportal` to avoid duplicating league-to-sport mapping
- `paper_bet` validates event is SCHEDULED before placing
- All error returns include `error_type` for debuggability

## Closes #271

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ htmlcov/
 # Config files and data
 *.csv
 *.json
+!.mcp.json
 
 # ESPN fixture data (regenerated via scripts/ingest_espn_fixtures.py)
 data/espn_fixtures/
--- a/.mcp.json
+++ b/.mcp.json
@@ -1,0 +1,9 @@
+{
+  "mcpServers": {
+    "odds": {
+      "type": "stdio",
+      "command": "uv",
+      "args": ["run", "fastmcp", "run", "packages/odds-mcp/odds_mcp/server.py"]
+    }
+  }
+}
--- a/packages/odds-cli/odds_cli/commands/copy_from_prod.py
+++ b/packages/odds-cli/odds_cli/commands/copy_from_prod.py
@@ -15,7 +15,7 @@ console = Console()
 def copy_from_production(
     start: str = typer.Argument(..., help="Start date (YYYY-MM-DD)"),
     end: str = typer.Argument(..., help="End date (YYYY-MM-DD)"),
-    sport: str = typer.Option("basketball_nba", "--sport", "-s", help="Sport to copy"),
+    sport: str = typer.Option("soccer_epl", "--sport", "-s", help="Sport to copy"),
     dry_run: bool = typer.Option(False, "--dry-run", help="Perform dry run without writing data"),
     skip_existing: bool = typer.Option(
         True, "--skip-existing/--overwrite", help="Skip existing events"
@@ -24,22 +24,15 @@ def copy_from_production(
     local_url: str | None = typer.Option(None, "--local-url", help="Local database URL"),
 ):
     """
-    Copy completed games from production database to local database.
-
-    This performs a selective merge, copying only completed games with results
-    while preserving existing local data.
+    Copy events, snapshots, and predictions from production to local database.
 
     Examples:
-        # Copy October-November 2025 data (dry run first)
-        odds copy from-prod 2025-10-23 2025-11-07 --dry-run
+        odds copy from-prod 2025-08-01 2026-04-10
 
-        # Actually copy the data
-        odds copy from-prod 2025-10-23 2025-11-07
+        odds copy from-prod 2025-08-01 2026-04-10 --dry-run
 
-        # Copy with overwrite
-        odds copy from-prod 2025-10-23 2025-11-07 --overwrite
+        odds copy from-prod 2025-08-01 2026-04-10 --overwrite
     """
-    # Validate dates
     try:
         datetime.strptime(start, "%Y-%m-%d")
         datetime.strptime(end, "%Y-%m-%d")
--- a/packages/odds-lambda/odds_lambda/copy_from_production.py
+++ b/packages/odds-lambda/odds_lambda/copy_from_production.py
@@ -1,8 +1,7 @@
 """
-Copy completed games from production database to local database.
+Copy data from production database to local database.
 
-This script performs a selective merge, copying only completed games with results
-from the production database while preserving existing local data.
+Replicates events, snapshots, and predictions for a given sport and date range.
 """
 
 from __future__ import annotations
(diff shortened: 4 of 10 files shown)


Acusick1 (commented):
## Review: odds-mcp FastMCP server (#273)

CI passes. Co-authored by Claude, so commenting only.

### Must-fix

**1. Wrong default retail bookmakers (bug)**
`_DEFAULT_RETAIL_BOOKMAKERS = ["fanduel", "draftkings", "betmgm"]` — these are US bookmakers that don't exist in your EPL OddsPortal data. Every EPL config uses UK bookmakers (`bet365`, `betway`, `betfred`, etc.). The `get_event_features` tool will silently produce NaN features for retail metrics against live data.

Should match the EPL configs, e.g.:
```python
_DEFAULT_RETAIL_BOOKMAKERS = ["bet365", "betway", "betfred"]
```

The `DEFAULT_RETAIL_BOOKMAKERS` in `feature_extraction.py` has the same problem (inherited from the NBA era), but the MCP server should default correctly for its primary use case.

**2. `get_current_odds` returns raw `raw_data` JSON blob**
The `get_odds_history` tool deliberately uses `extract_odds_from_snapshot()` to produce structured output (as called out in the PR description). But `get_current_odds` returns the full `raw_data` blob via `include_raw_data=True`. For a single event this is the most important tool to get right — an agent is most likely to call this one. Should use `extract_odds_from_snapshot()` here too, with `include_raw_data` as a fallback/optional param.

### Should-fix

**3. `get_event_features` only extracts tabular features**
Issue #271 scope says `get_event_features(event_id)` should return "feature vector (standings, schedule, stats)". The implementation only runs `TabularFeatureExtractor`. If the other feature groups aren't wired up yet, the docstring should say "tabular features only" rather than implying completeness. Not blocking since those features are still experimental, but worth calling out.

**4. `get_model_prediction` reads stored predictions, doesn't run inference**
Issue #271 success criteria: "get_model_prediction loads model and runs inference." The implementation queries the `Prediction` table for pre-scored results. This is arguably a better design (avoids cold-loading the model per-call), but the tool name and issue intent suggest on-demand inference. Either rename to `get_stored_predictions` or add a note in the docstring that this reads pre-scored predictions from the scoring job.

**5. `paper_bet` bare `except Exception`**
```python
try:
    trade = await place_trade(...)
    await session.commit()
except Exception as e:
    return {"error": str(e), "error_type": type(e).__name__}
```
`place_trade` raises `ValueError` for validation errors. Catching all exceptions here silently swallows DB errors (connection failures, constraint violations) that should probably propagate or at least log. Consider catching `ValueError` specifically and letting unexpected errors bubble up, or at minimum add a `logger.error` call in the except block (like `refresh_scrape` does).

### Nits

- `_DEFAULT_SHARP_BOOKMAKERS` duplicates the constant from `feature_extraction.py`. Consider importing `DEFAULT_SHARP_BOOKMAKERS` (and fixing the retail one there too), or at least adding a comment noting the duplication.
- `get_odds_history` has a lazy import of `extract_odds_from_snapshot` inside the function body, but `get_event_features` also lazy-imports from `odds_analytics`. These are already top-level dependencies of the package — no reason to defer them. Move to module-level imports for consistency.

### What looks good

- Clean thin-wrapper design — tools delegate to existing service code rather than reimplementing
- Good use of `LEAGUE_SPECS` for sport resolution instead of hardcoded mapping
- Proper `async with async_session_maker()` pattern throughout
- Event status validation before placing paper bets
- `_safe_float` handling for numpy NaN values
- Tests cover serialization helpers and key error paths
- `.mcp.json` for Claude Code discovery is clean

Acusick1 (commented):
All feedback addressed cleanly in the latest commit. LGTM.

- Retail bookmakers fixed to EPL defaults
- `get_current_odds` now uses structured extraction with opt-in raw_data
- `get_predictions` rename + accurate docstring
- `get_event_features` docstring scoped to tabular only
- `paper_bet` catches `ValueError` specifically, logs unexpected errors
- Lazy imports promoted to module-level

CI green. Ship it.

PR #280, opened 2026-04-11
Standardise job signatures with typed JobContext

### LINKED ISSUE #278: Standardise job signatures with typed JobContext
## Goal

The Lambda handler and local scheduler backend both use `inspect.signature` to filter kwargs before calling job functions. This is fragile and duplicated. A typed `JobContext` gives all jobs a consistent interface, makes required parameters explicit, and removes the introspection.

## Scope

- Define a `JobContext` dataclass in `odds_lambda/scheduling/jobs.py` with fields: `sport: str | None = None`, `retry_count: int = 0`
- Update all 10 job `main()` functions to accept `ctx: JobContext` as their only parameter
- Jobs that require specific fields (e.g. `fetch_oddsportal` requires `sport`) assert them: `assert ctx.sport is not None`
- Simplify `lambda_handler._run_job_async` to construct `JobContext` from the event payload and pass it directly — no `inspect.signature`
- Simplify local backend's `schedule_next_execution` to store `JobContext` kwargs — no `inspect.signature`
- Remove `**_kwargs` from all jobs

## Success criteria

- [ ] `JobContext` dataclass defined with `sport` and `retry_count` fields
- [ ] All 10 job `main()` functions accept `ctx: JobContext`
- [ ] Jobs that require `sport` assert it at the top
- [ ] No `inspect.signature` usage in handler or local backend
- [ ] Existing tests pass

## Location

- `packages/odds-lambda/odds_lambda/scheduling/jobs.py` — `JobContext` definition
- `packages/odds-lambda/odds_lambda/lambda_handler.py` — `_run_job_async`
- `packages/odds-lambda/odds_lambda/scheduling/backends/local.py` — `schedule_next_execution`
- `packages/odds-lambda/odds_lambda/jobs/` — all 10 job files

## Notes

- `backfill_polymarket.main()` currently takes `start_date` and `end_date` — these are manual invocation params, not scheduler params. They should move to `JobContext` or the job should parse them from the payload separately.

Acusick1:
## Summary
- Define `JobContext` dataclass in `odds_lambda/scheduling/jobs.py` with `sport`, `retry_count`, and job-specific fields (`include_spreads`, `include_totals`, `dry_run`, `lookback_hours`, `lookahead_hours`, `extra`)
- Update all 10 job `main()` functions to accept `ctx: JobContext` as their only parameter
- Simplify `lambda_handler._run_job_async` — construct `JobContext.from_payload()` and pass directly, removing `inspect.signature` introspection
- Simplify local backend's `schedule_next_execution` — pass `JobContext` to APScheduler instead of filtering kwargs via `inspect.signature`
- Update CLI callers and test files to use `JobContext`

## Closes #278

### DIFF
--- a/packages/odds-cli/odds_cli/commands/polymarket.py
+++ b/packages/odds-cli/odds_cli/commands/polymarket.py
@@ -86,6 +86,7 @@ async def _backfill_async(
 
     # Import job module
     from odds_lambda.jobs import backfill_polymarket
+    from odds_lambda.scheduling.jobs import JobContext
 
     # Execute with progress indicator
     with Progress(
@@ -97,9 +98,11 @@ async def _backfill_async(
 
         try:
             await backfill_polymarket.main(
-                include_spreads=include_spreads,
-                include_totals=include_totals,
-                dry_run=dry_run,
+                JobContext(
+                    include_spreads=include_spreads,
+                    include_totals=include_totals,
+                    dry_run=dry_run,
+                )
             )
             progress.update(task, completed=True)
 
--- a/packages/odds-cli/odds_cli/commands/scheduler.py
+++ b/packages/odds-cli/odds_cli/commands/scheduler.py
@@ -119,12 +119,19 @@ def run_once(
 
     try:
         # Use centralized job registry
-        from odds_lambda.scheduling.jobs import JobContext, get_job_function, list_available_jobs
+        from odds_lambda.scheduling.jobs import (
+            JobContext,
+            get_job_function,
+            list_available_jobs,
+            resolve_job_name,
+        )
 
-        job_func = get_job_function(job)
+        base_name, sport = resolve_job_name(job)
+        job_func = get_job_function(base_name)
+        ctx = JobContext(sport=sport) if sport else JobContext()
 
         # Run job
-        asyncio.run(job_func(JobContext()))
+        asyncio.run(job_func(ctx))
 
         console.print(f"\n[bold green]✓ {job} completed[/bold green]")
 
--- a/packages/odds-cli/odds_cli/commands/scheduler.py
+++ b/packages/odds-cli/odds_cli/commands/scheduler.py
@@ -60,8 +60,9 @@ def start_local():
 
         try:
             from odds_lambda.jobs import fetch_odds
+            from odds_lambda.scheduling.jobs import JobContext
 
-            await fetch_odds.main()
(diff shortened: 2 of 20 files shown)


Acusick1 (commented):
## Review: Standardise job signatures with typed JobContext

Clean refactor — the `inspect.signature` introspection was fragile and this is a much better design. CI passes. A few items:

### Bug: `run_once` doesn't extract sport from compound job name

`scheduler.py:run_once` calls `asyncio.run(job_func(JobContext()))` — an empty context. If you run `odds scheduler run-once fetch-odds:soccer_epl`, `get_job_function` resolves the base name but the sport suffix is lost. The lambda handler and local backend both correctly extract sport via `resolve_job_name` or `from_payload`, but `run_once` doesn't.

Should mirror what the local backend does:
```python
from odds_lambda.scheduling.jobs import JobContext, get_job_function, resolve_job_name, list_available_jobs

_, sport = resolve_job_name(job)
job_func = get_job_function(job)
ctx = JobContext(sport=sport) if sport else JobContext()
asyncio.run(job_func(ctx))
```

Same applies to the bootstrap call in `run_scheduler` — though that one falls back to config, so it's less critical.

### Nit: duplicate imports in test_scheduler_e2e.py

`JobContext` is imported at the top of the file (line 22 in the new version) but also re-imported locally inside `wrapped_fetch_odds` (line 199) and the tier scheduling test (line 325). The local imports are unnecessary.

### Nit: docstring dash style

Some docstrings changed `—` (em dash) to `--` (double hyphen) — e.g. `fetch_oddsportal.py`, `fetch_oddsportal_results.py`. Inconsistent with the rest of the codebase which uses em dashes. Not worth a separate commit, but worth keeping consistent if you're touching these lines anyway.

### Design question: `extra` dict

The `extra` catch-all dict is forward-compatible but nothing reads from it today. Is this intentional for future use (e.g. passing through unknown Lambda payload fields), or could it be dropped until needed? Not blocking — just want to understand the intent.

### Overall

The core change is solid. The `run_once` sport extraction is the only functional issue.

Acusick1 (commented):
All items addressed:

- ✅ `run_once` now extracts sport via `resolve_job_name` — bug fixed
- ✅ Duplicate imports removed from `test_scheduler_e2e.py`
- ✅ CI passes

LGTM. Can't approve since Claude co-authored.

PR #281, opened 2026-04-11
Centralized team name normalization

### LINKED ISSUE #185: Unified team name normalization across data sources
## Goal

OddsPortal and the Odds API use different team names for the same teams (e.g. "Wolves" vs "Wolverhampton Wanderers", "Tottenham" vs "Tottenham Hotspur"). If both sources cover the same sport, `find_or_create_event` fails to match on exact name comparison and creates duplicate events. Team names need a single normalization path that both pipelines share.

## Scope

- Add `normalize_team_name()` in `odds_core` with whitespace/case cleanup and an explicit alias map resolving abbreviations to canonical (Odds API) names
- Add Pydantic `@field_validator` on `Event.home_team` / `Event.away_team` so normalization is enforced at the model level
- Move `find_or_create_event` and `_build_event_id` from `fetch_oddsportal.py` into `OddsWriter` — event identity resolution is a storage concern, not job-specific logic
- Update `fetch_oddsportal.py` to use `writer.find_or_create_event()`
- Add Alembic migration to normalize existing team names in the `events` table
- Add tests for normalization, field validator, and the moved writer method

## Location

- `packages/odds-core/odds_core/team.py` — new module
- `packages/odds-core/odds_core/models.py` — field validator
- `packages/odds-lambda/odds_lambda/storage/writers.py` — `find_or_create_event` method
- `packages/odds-lambda/odds_lambda/jobs/fetch_oddsportal.py` — remove local matching logic

## Implementation approach

**Alias map in code** — a flat `dict[str, str]` mapping known abbreviations to canonical names. Title case applied before lookup so aliases only need one entry per variant. No fuzzy matching, no regex — explicit and debuggable.

**Canonical names** are the Odds API full names (most unambiguous, stable). Known EPL mismatches:

| OddsPortal | Canonical (Odds API) |
|---|---|
| Brighton | Brighton and Hove Albion |
| Leeds | Leeds United |
| Manchester Utd | Manchester United |
| Newcastle | Newcastle United |
| Nottingham | Nottingham Forest |
| Sheffield Utd | Sheffield United |
| Tottenham | Tottenham Hotspur |
| West Ham | West Ham United |
| Wolves | Wolverhampton Wanderers |

**Migration**: UPDATE `events` rows for each alias mapping (both `home_team` and `away_team`). Do not modify `raw_data` JSON in snapshots — that preserves the original source response.

## Success criteria

- [ ] `normalize_team_name("Wolves")` returns `"Wolverhampton Wanderers"` (and all other aliases)
- [ ] `normalize_team_name("  manchester   united ")` returns `"Manchester United"` (whitespace/case)
- [ ] `normalize_team_name("Arsenal")` returns `"Arsenal"` (passthrough for non-aliased names)
- [ ] Constructing `Event(home_team="Wolves", ...)` stores `"Wolverhampton Wanderers"`
- [ ] `find_or_create_event` no longer exists in `fetch_oddsportal.py` — lives on `OddsWriter`
- [ ] Existing OddsPortal events have normalized names after migration
- [ ] All existing tests pass

## Notes

- `team_abbrev()` in `oddsportal_common.py` operates on already-normalized names after this change (model validator runs on Event construction). No change needed to the function itself.
- The alias map will grow as new leagues are onboarded — keep it in code for now, consider config file if it becomes unwieldy.

Acusick1:
## Summary
- Add `odds_core.team.normalize_team_name()` — single source of truth for mapping source-specific team names (OddsPortal, FDUK, ESPN, FBref) to canonical short forms stored in the events table
- Normalize `Event.home_team` and `Event.away_team` on construction via `model_post_init`
- Move `find_or_create_event()` and `_build_event_id()` from `fetch_oddsportal.py` to `OddsWriter` — event identity resolution is a storage concern. Inputs are normalized before querying and ID generation
- Data migration normalizes existing team names in the events table (parameterized SQL, reversible)
- Delegate `scripts/team_names.py` to the core normalization function, removing duplicate alias dicts

## Test plan
- [x] Unit tests for `normalize_team_name()` (aliases, passthrough, whitespace, case)
- [x] Unit tests for `Event` model normalization via `model_post_init`
- [x] Unit tests for `OddsWriter.find_or_create_event()` (create, find existing, idempotency)
- [x] Test for normalization-during-query path (query with alias when DB has canonical name)
- [x] CI green (1651 passed)

Closes #185

### DIFF
--- a/migrations/versions/e7f2a1b3c4d5_normalize_team_names.py
+++ b/migrations/versions/e7f2a1b3c4d5_normalize_team_names.py
@@ -6,6 +6,7 @@ Create Date: 2026-04-10 12:00:00.000000
 
 """
 
+import sqlalchemy as sa
 from alembic import op
 
 # revision identifiers, used by Alembic.
@@ -44,7 +45,8 @@ def upgrade() -> None:
     for alias, canonical in _ALIAS_MAP.items():
         for col in ("home_team", "away_team"):
             op.execute(
-                f"UPDATE events SET {col} = '{canonical}' WHERE {col} = '{alias}'"  # noqa: S608
+                sa.text(f"UPDATE events SET {col} = :val WHERE {col} = :old"),
+                {"val": canonical, "old": alias},
             )
 
 
@@ -75,5 +77,6 @@ def downgrade() -> None:
     for canonical, alias in _REVERSE.items():
         for col in ("home_team", "away_team"):
             op.execute(
-                f"UPDATE events SET {col} = '{alias}' WHERE {col} = '{canonical}'"  # noqa: S608
+                sa.text(f"UPDATE events SET {col} = :val WHERE {col} = :old"),
+                {"val": alias, "old": canonical},
             )
--- a/migrations/versions/e7f2a1b3c4d5_normalize_team_names.py
+++ b/migrations/versions/e7f2a1b3c4d5_normalize_team_names.py
@@ -45,8 +45,9 @@ def upgrade() -> None:
     for alias, canonical in _ALIAS_MAP.items():
         for col in ("home_team", "away_team"):
             op.execute(
-                sa.text(f"UPDATE events SET {col} = :val WHERE {col} = :old"),
-                {"val": canonical, "old": alias},
+                sa.text(f"UPDATE events SET {col} = :val WHERE {col} = :old").bindparams(
+                    val=canonical, old=alias
+                )
             )
 
 
@@ -77,6 +78,7 @@ def downgrade() -> None:
     for canonical, alias in _REVERSE.items():
         for col in ("home_team", "away_team"):
             op.execute(
-                sa.text(f"UPDATE events SET {col} = :val WHERE {col} = :old"),
-                {"val": alias, "old": canonical},
+                sa.text(f"UPDATE events SET {col} = :val WHERE {col} = :old").bindparams(
+                    val=alias, old=canonical
+                )
             )
--- a/migrations/versions/e7f2a1b3c4d5_normalize_team_names.py
+++ b/migrations/versions/e7f2a1b3c4d5_normalize_team_names.py
@@ -1,0 +1,79 @@
+"""normalize team names in events table
+
+Revision ID: e7f2a1b3c4d5
+Revises: 021f43706c10
+Create Date: 2026-04-10 12:00:00.000000
+
+"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "e7f2a1b3c4d5"
+down_revision = "021f43706c10"
+branch_labels = None
+depends_on = None
+
+# Alias -> canonical (same mappings as odds_core.team._TEAM_ALIASES).
+# Only includes names that could exist in the events table.
+_ALIAS_MAP: dict[str, str] = {
+    "AFC Bournemouth": "Bournemouth",
+    "Brighton & Hove Albion": "Brighton",
+    "Brighton and Hove Albion": "Brighton",
+    "Cardiff City": "Cardiff",
+    "Huddersfield Town": "Huddersfield",
+    "Hull City": "Hull",
+    "Ipswich Town": "Ipswich",
+    "Leicester City": "Leicester",
+    "Leeds United": "Leeds",
+    "Manchester United": "Manchester Utd",
+    "Newcastle United": "Newcastle",
+    "Norwich City": "Norwich",
+    "Nottingham Forest": "Nottingham",
+    "Sheffield United": "Sheffield Utd",
+    "Stoke City": "Stoke",
+    "Swansea City": "Swansea",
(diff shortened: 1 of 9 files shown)


Acusick1:
## Design decision: canonical name direction

The issue spec says canonical names should be Odds API long forms (e.g., `normalize_team_name("Wolves")` → `"Wolverhampton Wanderers"`). This PR does the **reverse** — canonical names are the short forms already in the DB (e.g., "Wolves", "Tottenham", "Manchester Utd").

**Rationale:**
- The existing events table already uses short forms (from The Odds API responses, which actually use short names like "Wolves" not "Wolverhampton Wanderers")
- Avoids a mass data migration to longer names
- Both sources still converge to the same canonical form, so deduplication works correctly

**How deduplication works across sources:**
1. Odds API → `create_scheduled_event()` constructs `Event(home_team="Wolves")` → `model_post_init` passes through (already canonical)
2. OddsPortal → `writer.find_or_create_event(home_team="Wolverhampton Wanderers")` → normalizes to "Wolves" before querying → matches existing Odds API event

@Acusick1 confirming this direction is acceptable before merge.

Acusick1 (commented):
## Review: Centralized team name normalization

**CI:** Passing (1651 tests). Co-authored by Claude — commenting only, not approving.

### Overall

Clean consolidation. Single source of truth in `odds_core.team`, model-level enforcement via `model_post_init`, migration for existing data, and the `find_or_create_event` move to `OddsWriter` is a good layering call. A few things worth addressing:

---

### Issues

**1. Event ID format change (silent breaking change)**

`_build_event_id` now receives **normalized** names (short forms) instead of raw source names. This changes IDs for future OddsPortal events:

- Before: `_build_event_id("Wolverhampton Wanderers", ...)` → `op_live_WOLWAN_ARS_...`
- After: `_build_event_id("Wolves", ...)` → `op_live_WOL_ARS_...`

The team+date query in `find_or_create_event` means existing events will still be found correctly (post-migration). But any code or external references that construct or parse event IDs will break. Worth calling out explicitly — is anything downstream sensitive to the ID format?

**2. `scripts/team_names.py` — compatibility artifacts**

The `source` parameter is now dead code (always ignored). The `__all__` with "Re-export for backward compatibility" comment is the kind of compatibility hack the project conventions discourage. Since all callers are in-repo scripts, just remove the `source` param and the `__all__`:

```python
def normalize_team(name: str) -> str:
    return normalize_team_name(name)
```

Update the three callers that pass `source=` (only `ingest_fpl_availability.py` does, and the FPL-specific alias `"Man Utd"` is now handled by the core function's case-insensitive lookup).

**3. `team_abbrev` import in OddsWriter**

`OddsWriter` (generic storage) now imports `team_abbrev` from `oddsportal_common` (source-specific). The `_build_event_id` method is inherently OddsPortal-specific (`op_live_` prefix), so it's a bit odd on the generic writer. Not a blocker for this PR, but worth noting — if another source needs its own ID scheme, this method will need to be polymorphic or extracted.

### Nits

- **Double normalization**: `find_or_create_event` normalizes explicitly, then `Event()` constructor normalizes again via `model_post_init`. Harmless but redundant — could add a comment noting the belt-and-suspenders intent, or skip one layer.
- **Migration alias map duplication**: Correct and necessary (migrations must be self-contained). No action needed.

### Looks good

- `model_post_init` is the right hook for SQLModel `table=True` — good that you iterated through `__init__`, `@field_validator`, and `@validates` to land here
- Case-insensitive lookup with lowercase keys is clean
- Parameterized SQL in migration (`sa.text().bindparams()`) handles apostrophes correctly
- Test coverage is solid — alias lookup, passthrough, whitespace, Event construction, and the critical normalization-during-query path
- Downgrade path is lossy but documented

Acusick1:
Addressed the review findings:

**1. Event ID format change** — Confirmed: nothing downstream parses or constructs event IDs. They're opaque keys used only for DB lookups. The `find_or_create_event` team+date query finds existing events regardless of ID format, and the migration doesn't change existing IDs. New OddsPortal events will get shorter abbreviations (e.g., `op_live_WOL_ARS_...` instead of `op_live_WOLWAN_ARS_...`) but this is cosmetic.

**2. `scripts/team_names.py` compatibility artifacts** — Fixed in 96817f2. Removed the dead `source` parameter, the `__all__` re-export, and updated the FPL caller.

**3. `team_abbrev` import in OddsWriter** — Acknowledged. `_build_event_id` is inherently OddsPortal-specific (`op_live_` prefix). If another source needs its own ID scheme this will need to be extracted, but that's out of scope for this PR.

**Nit (double normalization)** — Intentional belt-and-suspenders: `find_or_create_event` normalizes for the query, `model_post_init` normalizes as a safety net for any other Event construction path.

Acusick1 (commented):
Previous findings addressed — `source` param and `__all__` removed, FPL caller updated. CI green. LGTM.

PR #283, opened 2026-04-11
Fix stale event settlement and clean up duplicate OddsPortal events

### LINKED ISSUE #279: Fix stale event settlement and clean up duplicate OddsPortal events
## Goal

19 `op_live_*` EPL events in prod are stuck — 6 SCHEDULED, 13 LIVE — because `fetch_oddsportal_results` only queries SCHEDULED status and team name mismatches prevent matching (see #185). Additionally, 16 of ~19 future EPL matches exist as both an Odds API event and an `op_live_*` duplicate. After #185 lands (preventing new duplicates), this issue cleans up existing data and fixes the LIVE settlement gap.

## Scope

- Fix `get_pending_events()` in `fetch_oddsportal_results.py` to include `EventStatus.LIVE` alongside `SCHEDULED`
- Add one-time data migration to merge duplicate events: for each `op_live_*` event that has a matching Odds API event (same match, ±24h window), reassign its `OddsSnapshot` rows to the API event ID and delete the `op_live_*` event
- Settle remaining stale events (those without an API counterpart) using `fetch_oddsportal_results` or manual status update
- Remove "Short fixture lookahead" item from `TODO.md` — prod has 38 future events through April 20, not a real issue

## Location

- `packages/odds-lambda/odds_lambda/jobs/fetch_oddsportal_results.py` — `get_pending_events()`
- Alembic migration for duplicate merging

## Constraints

- Depends on #185 (normalization must be in place before dedup migration runs, otherwise the team name matching in the migration would need its own normalization logic)
- Migration must preserve all `OddsSnapshot` data — reassign `event_id`, never delete snapshots
- Migration must handle the case where both events have snapshots (merge, don't overwrite)

## Success criteria

- [ ] `get_pending_events()` returns events with status SCHEDULED or LIVE
- [ ] After migration: no `op_live_*` events that duplicate an Odds API event for the same match
- [ ] After migration: all past EPL events are FINAL (no stale SCHEDULED/LIVE)
- [ ] No `OddsSnapshot` rows lost during dedup
- [ ] "Short fixture lookahead" removed from TODO.md

## Notes

- The 83 stale NBA API events are from deprioritized NBA fetching — out of scope, but could be bulk-settled to FINAL in the same migration if convenient
- `op_live_*` events that have no Odds API counterpart (e.g. `op_live_CRYPAL_NEW_2026-04-11`) should be kept — they're the only record of that match

Acusick1:
## Summary

- Fix `get_pending_events()` to query both `SCHEDULED` and `LIVE` events — 13 LIVE EPL events were stuck and never settled
- Add one-time Alembic data migration that:
  - Matches `op_live_*` duplicate events to their Odds API counterparts (same teams, ±24h window) and reassigns all child rows (snapshots, odds, predictions, paper_trades, data_quality_logs) before deleting the duplicate
  - Bulk-settles stale past EPL SCHEDULED/LIVE events to FINAL
  - Bulk-settles stale NBA API events to FINAL (convenient cleanup)
- Add tests for `get_pending_events()` covering SCHEDULED, LIVE, FINAL, and future event filtering
- Remove resolved TODO items (stale events, duplicates, #185)

## Closes #279


Acusick1 (commented):
## Review: Fix stale event settlement and clean up duplicate OddsPortal events

**CI:** Passing. **Co-authored by Claude — commenting only.**

### Overall

Clean, well-scoped PR. The `get_pending_events()` fix is correct and the tests cover the right cases. The migration is straightforward and well-structured. A couple of items worth addressing:

### Issue: Missing child tables in migration

`_CHILD_TABLES` covers `odds_snapshots`, `odds`, `predictions`, `paper_trades`, `data_quality_logs` — but three other models also have `event_id` FK to `events`:

- `polymarket_events` (nullable FK)
- `nba_team_game_logs`
- `nba_injury_reports`

In practice these are all NBA/Polymarket tables and won't reference `op_live_*` EPL events, so the DELETE would succeed. But if any row *did* exist, you'd get an FK constraint violation at migration time. Worth adding them to `_CHILD_TABLES` for completeness, or at minimum adding a comment noting they're excluded intentionally because they're NBA-only.

### Nit: Stale docstring in source

The diff updates the `get_pending_events` docstring to say "SCHEDULED or LIVE" — good. Just confirming the module-level docstring at line 7 was also updated (the diff shows it was). 👍

### Migration design notes (non-blocking)

- `downgrade()` is a no-op with a print — reasonable for a data-only migration, but consider raising `NotImplementedError` so an accidental `alembic downgrade` fails loudly rather than silently succeeding and leaving the DB in an inconsistent state.
- `NOW()` in the bulk-settle queries is fine for a one-time migration. Just noting that `updated_at` won't use the app-level `datetime.now(UTC)` pattern, but that's expected for raw SQL.

### Checklist

- [x] CI passes
- [x] `get_pending_events` correctly includes LIVE status
- [x] Tests cover SCHEDULED, LIVE, FINAL, and future event filtering
- [x] Migration handles child row reassignment before DELETE (correct FK order)
- [x] `data_quality_logs` included in child tables (second commit fix)
- [x] f-string table interpolation is from hardcoded constant, not user input — safe
- [x] No secrets, no injection risks
- [x] Type hints present on all modified functions

Acusick1 (commented):
## Review of #283

CI passes. Claude co-authored, so commenting only.

### Migration (`d8858af1ee85`)

**Looks good overall.** The three-step approach (merge duplicates → settle EPL → settle NBA) is clean, child table list is complete, and `downgrade()` correctly raises.

One concern:

1. **Unique constraint violations on child row reassignment.** If both the `op_live_*` event and the API event have rows in a child table with a unique constraint that includes `event_id` (e.g. `predictions` has a unique on `(event_id, snapshot_id, model_name)`), the `UPDATE ... SET event_id = :new` could violate that constraint and abort the migration. This is unlikely given the data (OP events probably don't have predictions), but worth thinking about. A `ON CONFLICT DO NOTHING` approach via raw SQL or a pre-check would be safer if any of these tables have unique constraints involving `event_id`.

### `get_pending_events()` fix

Clean, correct. The `.in_([SCHEDULED, LIVE])` change matches the issue requirement. Docstring on master still says "SCHEDULED events" (line 70) — the diff updates it on the branch, confirmed.

### Tests

Good coverage of the four key states (SCHEDULED, LIVE, FINAL, future). Tests use the `test_session` fixture for real DB interaction. Clean and focused.

### Minor nits

- The PR description mentions "Remove resolved TODO items" but TODO.md isn't in the changed files — looks like that was dropped in a subsequent commit. No issue, just noting the description is slightly stale.
- Migration `Create Date` is hardcoded (`2026-04-11 12:00:00.000000`) rather than auto-generated — cosmetic only, doesn't affect anything.

### Summary

The core logic is correct and well-tested. The only substantive concern is potential unique constraint violations during child row reassignment in the migration — worth verifying against prod data before running.

Acusick1 (commented):
## Re-review of #283 (after 4th commit)

CI passes. Previous concern about unique constraint violations on `predictions` has been addressed with `_reassign_predictions()` — correctly deletes conflicting rows (same snapshot_id + model_name) before moving the rest. Verified that `predictions` is the only child table with a unique constraint involving `event_id`.

**LGTM.** No remaining concerns. Clean split between simple UPDATE tables and the predictions special case. Ready to merge.

PR #308, opened 2026-04-14
Decouple scraper from MCP server via background worker job queue

### LINKED ISSUE #307: Decouple scraper from MCP server via background worker job queue
## Goal

The `refresh_scrape` MCP tool calls `ingest_league()` in-process, which launches Playwright/Chromium. This corrupts the stdio transport and kills the MCP server. Isolate scraping into a separate OS process by using the database as a job queue — the MCP server enqueues, a worker process executes.

## Scope

- Add `ScrapeJob` SQLModel (`scrape_jobs` table) with status enum (`pending`/`running`/`completed`/`failed`), league, market, timestamps, and flat `IngestionStats` result columns
- Rewrite `refresh_scrape` MCP tool to insert a `pending` row and return the job ID immediately — remove the `ingest_league` import from `server.py`
- Add `get_scrape_status` MCP tool to query job status and results by ID
- Add `odds worker start` CLI command with a polling loop that claims pending jobs, runs `ingest_league()`, and writes results back
- Add alembic migration for the new table

## Location

- `packages/odds-core/odds_core/scrape_job_models.py` (new — model + enum)
- `packages/odds-mcp/odds_mcp/server.py` (rewrite `refresh_scrape`, add `get_scrape_status`)
- `packages/odds-cli/odds_cli/commands/worker.py` (new — worker command)
- `packages/odds-cli/odds_cli/main.py` (register worker command)
- `migrations/env.py` (register model import)

## Implementation approach

- `ScrapeJob` follows the `PaperTrade` model pattern (`paper_trade_models.py`): SQLModel with `table=True`, `Field()` types, flat columns for result stats rather than JSON
- Worker claims jobs atomically: `UPDATE scrape_jobs SET status='running' WHERE id=X AND status='pending'` to avoid double-processing
- Worker builds a `LeagueSpec` from the job's league/market using `LEAGUE_SPECS` lookup, then calls `ingest_league(spec)` — same path the current tool uses
- `--once` flag on the worker for single-job-and-exit (testing). `--poll-interval` for the loop sleep (default 5s)
- In production, the existing scraper Lambda can check the `scrape_jobs` table for pending rows before/after its scheduled scrape — no new Lambda function needed

## Success criteria

- [ ] `refresh_scrape` returns a job ID immediately without launching Playwright
- [ ] MCP server stays alive after `refresh_scrape` is called
- [ ] `get_scrape_status` returns `pending`/`running`/`completed`/`failed` with stats on completion
- [ ] `odds worker start --once` picks up a pending job, scrapes, and writes results
- [ ] Existing MCP tests pass
- [ ] Alembic migration applies and downgrades cleanly

## Open questions

- Stale job cleanup: should the worker expire `running` jobs that have been stuck for N minutes (Lambda timeout, crash)?
- Should `refresh_scrape` check for an existing pending/running job for the same league+market and return that ID instead of creating a duplicate?

Acusick1:
## Summary

- Add `ScrapeJob` SQLModel with status enum (`pending`/`running`/`completed`/`failed`), league/market params, timestamps, and flat `IngestionStats` result columns
- Rewrite `refresh_scrape` MCP tool to insert a `pending` row and return job ID immediately — no more in-process Playwright that kills the stdio transport
- Deduplicates: returns existing pending/running job ID for the same league+market instead of creating duplicates
- Add `get_scrape_status` MCP tool to query job status and results by ID
- Add `odds worker start` CLI command with polling loop that claims pending jobs via `SELECT FOR UPDATE SKIP LOCKED`, runs `ingest_league()`, and writes results back
- Worker supports `--once` (single job and exit) and `--poll-interval` (default 5s), plus stale job expiry (>10 min running)
- Extract shared `LEAGUE_SPEC_BY_NAME` dict to `fetch_oddsportal.py`, use `dataclasses.replace()` for `LeagueSpec` construction
- 13 tests covering refresh_scrape (create, dedup, invalid league), get_scrape_status (all statuses), and worker claiming (FIFO, stale expiry)
- Alembic migration for `scrape_jobs` table with partial index on pending jobs

## Closes #307

### DIFF
--- a/migrations/env.py
+++ b/migrations/env.py
@@ -29,6 +29,7 @@ from odds_core.polymarket_models import (  # noqa: F401
     PolymarketPriceSnapshot,
 )
 from odds_core.prediction_models import Prediction  # noqa: F401
+from odds_core.scrape_job_models import ScrapeJob  # noqa: F401
 from sqlalchemy import pool
 from sqlalchemy.engine import Connection
 from sqlalchemy.ext.asyncio import async_engine_from_config
--- a/migrations/versions/5eca7168d060_add_scrape_jobs_table.py
+++ b/migrations/versions/5eca7168d060_add_scrape_jobs_table.py
@@ -25,7 +25,7 @@ def upgrade() -> None:
         sa.Column("market", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
         sa.Column(
             "status",
-            sa.Enum("pending", "running", "completed", "failed", name="scrapejobstatus"),
+            sa.String(20),
             nullable=False,
         ),
         sa.Column("created_at", sa.DateTime(timezone=True), nullable=False),
@@ -61,4 +61,3 @@ def downgrade() -> None:
     op.drop_index("ix_scrape_jobs_league_status", table_name="scrape_jobs")
     op.drop_index(op.f("ix_scrape_jobs_league"), table_name="scrape_jobs")
     op.drop_table("scrape_jobs")
-    sa.Enum("pending", "running", "completed", "failed", name="scrapejobstatus").drop(op.get_bind())
--- a/migrations/versions/5eca7168d060_add_scrape_jobs_table.py
+++ b/migrations/versions/5eca7168d060_add_scrape_jobs_table.py
@@ -1,0 +1,64 @@
+"""add scrape_jobs table
+
+Revision ID: 5eca7168d060
+Revises: 771d5b1b451c
+Create Date: 2026-04-14 22:29:02.211432
+
+"""
+
+import sqlalchemy as sa
+import sqlmodel
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "5eca7168d060"
+down_revision = "771d5b1b451c"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "scrape_jobs",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("league", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.Column("market", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.Column(
+            "status",
+            sa.Enum("pending", "running", "completed", "failed", name="scrapejobstatus"),
+            nullable=False,
+        ),
+        sa.Column("created_at", sa.DateTime(timezone=True), nullable=False),
+        sa.Column("started_at", sa.DateTime(timezone=True), nullable=True),
+        sa.Column("completed_at", sa.DateTime(timezone=True), nullable=True),
+        sa.Column("matches_scraped", sa.Integer(), nullable=True),
+        sa.Column("matches_converted", sa.Integer(), nullable=True),
+        sa.Column("events_matched", sa.Integer(), nullable=True),
+        sa.Column("events_created", sa.Integer(), nullable=True),
+        sa.Column("snapshots_stored", sa.Integer(), nullable=True),
+        sa.Column("error_message", sqlmodel.sql.sqltypes.AutoString(), nullable=True),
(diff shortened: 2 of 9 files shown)


Acusick1 (commented):
## Review: Decouple scraper from MCP server via background worker job queue

Clean design — the DB-as-queue pattern with `SELECT FOR UPDATE SKIP LOCKED` is the right call here. Dedup on existing pending/running jobs and stale job expiry are good additions that answer both open questions from the issue. The `dataclasses.replace()` cleanup and shared `LEAGUE_SPEC_BY_NAME` are welcome.

### CI blocker: Postgres enum in partial index

All 166 CI errors are the same root cause. The partial index uses `text("status = 'pending'")`, but when tests run `SQLModel.metadata.create_all` (rather than Alembic migrations), PostgreSQL sees the `status` column as enum type `scrapejobstatus` and rejects the uncast string literal:

```
invalid input value for enum scrapejobstatus: "pending"
[SQL: CREATE INDEX ix_scrape_jobs_pending ON scrape_jobs (status, created_at) WHERE status = 'pending']
```

**Fix**: cast the literal in both places (model + migration):
```python
postgresql_where=text("status = 'pending'::scrapejobstatus")
```

This affects `scrape_job_models.py:69` and `migrations/versions/5eca7168d060_add_scrape_jobs_table.py:52`.

### No tests for new code

The worker command and both MCP tool rewrites (`refresh_scrape`, `get_scrape_status`) have no tests. The issue's success criteria include "existing MCP tests pass" but there don't appear to be any tests covering:
- `refresh_scrape` enqueue + dedup logic
- `get_scrape_status` response shapes for each status
- Worker claim/process/fail lifecycle
- Stale job expiry

At minimum, the MCP tool logic is pure DB operations and straightforward to unit test.

### Minor observations

1. **Worker `_process_job` does two DB round-trips for success**: one to load the job (line 109), one to write results (line 133-144). Could claim league/market from the find query and do a single `UPDATE ... SET` on completion instead of select-mutate-commit. Not a blocker — just a note for if you want to tighten it.

2. **`_LEAGUE_SPEC_BY_NAME` alias in server.py**: Line 44 (`_LEAGUE_SPEC_BY_NAME = LEAGUE_SPEC_BY_NAME`) re-aliases the import as a private module var. The existing references all use `_LEAGUE_SPEC_BY_NAME`, so this avoids a larger diff, but it's a bit odd. Could just use the import directly. Minor.

3. **No graceful shutdown**: The worker `while True` loop has no signal handling. `Ctrl+C` will kill it mid-scrape, leaving a job in `running` state (caught by stale expiry, but 10 min delay). Fine for now given the stale expiry safety net, but worth noting if this becomes long-running in production.

### Summary

Fix the enum cast (CI blocker), consider adding basic tests for the new MCP tools. Architecture is sound.

Acusick1 (commented):
## Follow-up review

Tests look good — solid coverage of the MCP tools (create, dedup pending/running/completed, different market, all status response shapes) and worker claiming (oldest first, stale expiry, recent not expired). CI is green now.

### Migration/model mismatch (needs fix)

The model was changed to `native_enum=False` (VARCHAR storage) to fix the partial index issue, but the **migration still creates a native PostgreSQL enum**:

```python
# migration line 28 — creates native enum type
sa.Enum("pending", "running", "completed", "failed", name="scrapejobstatus")
```

```python
# model — expects VARCHAR(20)
SAEnum(ScrapeJobStatus, native_enum=False, length=20)
```

This means:
- **Tests** (`metadata.create_all`) create a VARCHAR column — CI passes
- **Production** (Alembic migration) creates a native enum column — the original partial index bug resurfaces, and the ORM/schema disagree

The migration needs to match the model: replace `sa.Enum(...)` with `sa.String(20)` and remove the `sa.Enum(...).drop()` call in downgrade.

Everything else looks good.

Acusick1 (commented):
Migration fixed — `sa.String(20)` matches the model, enum drop removed from downgrade. CI green. All prior feedback addressed. LGTM.

PR #311, opened 2026-04-15
Add time-windowed sharp price lookup across recent snapshots

### LINKED ISSUE #309: Add time-windowed sharp price lookup across recent snapshots
## Goal

The sharp-soft spread tool, match brief stamping, and feature display all rely on a single latest snapshot for sharp bookmaker prices. If that snapshot's scrape didn't capture Betfair Exchange (or Pinnacle), the sharp reference is missing entirely — even when a nearby snapshot has it. This creates a data availability gap that undermines the agent's ability to identify mispricings.

## Scope

- Add `get_sharp_prices(event_id, market, sharp_bookmakers, lookback_hours)` method to `OddsReader` that scans snapshots within a time window (newest-first) and returns the best available sharp price per outcome
- Update `get_sharp_soft_spread` to use the new reader method instead of `_snapshot_sharp_prices` on a single snapshot
- Update `save_match_brief` sharp price stamping to use the new reader method
- Add `sharp_lookback_hours` parameter to `get_sharp_soft_spread` tool with a defensive default (e.g. 2h) — short enough that a sharp price from a prior snapshot still reflects the same market conditions, avoiding misleading divergence calculations when the market has moved significantly between snapshots
- Remove `_snapshot_sharp_prices` from server.py if no longer used

## Location

- `packages/odds-lambda/odds_lambda/storage/readers.py` — new method on `OddsReader`
- `packages/odds-mcp/odds_mcp/server.py` — update `get_sharp_soft_spread`, `save_match_brief`

## Implementation approach

- The reader method queries `OddsSnapshot` rows for the event within `[now - lookback_hours, now]`, ordered by `snapshot_time DESC`
- For each outcome, iterate snapshots newest-first and stop at the first hit for a sharp bookmaker (respecting priority order of `sharp_bookmakers` list)
- Return a `SharpPriceMap` (existing type) plus metadata about which snapshot the price came from and its age
- Retail prices in `get_sharp_soft_spread` still come from the single latest snapshot — only the sharp reference uses the lookback
- `get_event_features` is intentionally **not** updated: `TabularFeatureExtractor` needs internally consistent snapshot data, and XGBoost handles NaN natively for missing sharp odds

## Success criteria

- [ ] `get_sharp_soft_spread` returns Betfair Exchange sharp prices when they exist in a recent (non-latest) snapshot
- [ ] `save_match_brief` stamps sharp prices found via lookback
- [ ] Sharp price response includes source snapshot time so the agent can judge staleness
- [ ] Default lookback is defensive — short enough to avoid comparing prices across a market move (2h aligns with closing-tier scrape frequency of 30min)
- [ ] Existing tests pass; new tests cover the lookback logic (sharp found in older snapshot, no sharp in window, expired window)

## Open questions

- Exact default for `sharp_lookback_hours` — 2h keeps it within ~4 scrape cycles at closing frequency. Could argue for shorter (1h) closer to kickoff when markets move fastest
- Whether `save_match_brief` should also expose `sharp_lookback_hours` as a parameter or just use the default

Acusick1:
## Summary

- Add `OddsReader.get_sharp_prices()` method that scans snapshots within a configurable time window (newest-first) to find sharp bookmaker prices, with per-outcome priority fallback
- New `SharpPriceMeta` / `SharpPriceResult` dataclasses carry provenance (snapshot ID, time, age) so the agent can judge staleness
- `get_sharp_soft_spread` now uses the lookback (default 2h, configurable via `sharp_lookback_hours` param) — retail prices still come from latest snapshot only
- `save_match_brief` uses the lookback (24h window) for sharp price stamping
- Removed `_snapshot_sharp_prices` helper from server.py (logic now lives in `OddsReader`)
- Added `odds-analytics` as explicit dependency of `odds-lambda`

## Closes #309


Acusick1 (commented):
## Review: PR #311 — Time-windowed sharp price lookup

**CI:** ✅ passing

**Cannot approve** — Claude is co-author. Commenting only.

### Overall

Clean, well-motivated change. The lookback approach correctly handles the real-world problem where a sharp bookmaker may be absent from the latest scrape but present in a nearby one. The implementation is straightforward (newest-first scan with per-outcome short-circuit) and well-tested.

### Checklist

| Area | Status |
|------|--------|
| Type hints | ✅ Complete |
| Async correctness | ✅ |
| Timezone handling | ✅ `datetime.now(UTC)` with injectable `now` |
| Tests | ✅ Good coverage — happy path, empty window, expired window, recency-vs-priority, e2e spread, brief stamping |
| N+1 queries | ✅ Single query per call |
| Breaking changes | ⚠️ See below |
| Security | ✅ No issues |

### Notes

1. **Dependency fix is correct.** `odds-lambda` already imported from `odds-analytics` in 3 files (`score_predictions.py`, `model_loader.py`, `paper_trading.py`) without declaring the dependency. This PR makes it explicit — good housekeeping.

2. **MCP tool contract change** — `get_sharp_soft_spread` response now includes `snapshot_time` and `age_seconds` in the sharp block, and gains a `sharp_lookback_hours` parameter. All additive (default-valued param, new fields in response), so existing callers won't break, but the agent prompts may want to reference the new staleness metadata.

3. **Minor: no early exit in `get_sharp_prices`** — once all outcomes are resolved, the loop continues scanning remaining snapshots (skipping via `continue`, but still calling `extract_odds_from_snapshot`). For h2h (2-3 outcomes) with a 2h window (1-5 snapshots), this is negligible. Not worth changing now, but if this ever extends to markets with many outcomes or wider windows, an early break would help.

4. **`test_no_sharp_returns_null_sharp`** — the 0.01h (36s) lookback window is tight but correct: `now` is anchored on the latest snapshot time, so `snap_new` is at the boundary and included, but it only has retail bookmakers. Assertion that all sharp fields are null is valid. The `if spread is not None` guard is slightly surprising — under what condition would spread be None here? If it can't happen, the guard masks potential regressions.

Looks good to merge. 👍

PR #312, opened 2026-04-15
Separate 1x2 and h2h market keys to prevent odds conflation

Acusick1:
## Summary

- **Problem**: 3-way (1x2) and 2-way (h2h) moneyline odds were both stored under `market_key="h2h"`. When an MLB `1x2` scrape job ran, regulation-time 3-way odds (with Draw outcomes at +700/+900) were stored alongside full-game 2-way moneyline data with no way to distinguish them at query time.
- **Fix**: 1x2 write paths now use `market_key="1x2"`. A data migration relabels 82K EPL + 171 MLB existing rows in both the `odds` table and `raw_data` JSON. Read paths (`extract_devigged_h2h_probs`, backtesting, arb strategy) accept both market keys where appropriate.
- EPL experiment configs updated from `markets: [h2h]` to `markets: [1x2]`.

## Changed files

**Write paths** (3 files): `oddsportal_adapter.py`, `fetch_oddsportal_results.py`, `ingest_oddsportal.py`
**Read paths** (4 files): `sequence_loader.py`, `feature_groups.py`, `backtesting/services.py`, `strategies.py`
**Migration**: `920aee156d9c` — relabels odds rows + patches raw_data JSON where Draw outcomes exist
**Configs**: 12 EPL experiment YAML files
**Tests**: `test_three_way_outcomes.py` fixtures updated to use `"1x2"`

## Test plan

- [x] All 168 unit tests pass (1 pre-existing failure in `test_settings_defaults`)
- [x] Migration verified: zero Draw outcomes remain under `market_key="h2h"`
- [x] EPL raw_data JSON patched correctly (`market_key: "1x2"`)
- [ ] Run EPL training pipeline with updated configs to verify feature extraction
- [ ] Run production migration on remote DB

### DIFF
--- a/experiments/configs/xgboost_epl_baseline_tuning.yaml
+++ b/experiments/configs/xgboost_epl_baseline_tuning.yaml
@@ -54,7 +54,7 @@ training:
     - betfred
     - bwin
     markets:
-    - h2h
+    - 1x2
     outcome: home
     closing_tier: closing
     sampling:
--- a/experiments/configs/xgboost_epl_betfair_sharp.yaml
+++ b/experiments/configs/xgboost_epl_betfair_sharp.yaml
@@ -55,7 +55,7 @@ training:
     - betfred
     - bwin
     markets:
-    - h2h
+    - 1x2
     outcome: home
     closing_tier: closing
     closing_source_priority:
--- a/experiments/configs/xgboost_epl_combined_all_tuning.yaml
+++ b/experiments/configs/xgboost_epl_combined_all_tuning.yaml
@@ -59,7 +59,7 @@ training:
     - betfred
     - bwin
     markets:
-    - h2h
+    - 1x2
     outcome: home
     closing_tier: closing
     closing_source_priority:
--- a/experiments/configs/xgboost_epl_combined_bet365sharp_tuning.yaml
+++ b/experiments/configs/xgboost_epl_combined_bet365sharp_tuning.yaml
@@ -55,7 +55,7 @@ training:
     - betfred
     - bwin
     markets:
-    - h2h
+    - 1x2
     outcome: home
     closing_tier: closing
     closing_source_priority:
--- a/experiments/configs/xgboost_epl_combined_match_stats_tuning.yaml
+++ b/experiments/configs/xgboost_epl_combined_match_stats_tuning.yaml
@@ -59,7 +59,7 @@ training:
     - betfred
     - bwin
     markets:
-    - h2h
+    - 1x2
     outcome: home
     closing_tier: closing
     closing_source_priority:
--- a/experiments/configs/xgboost_epl_combined_schedule_tuning.yaml
+++ b/experiments/configs/xgboost_epl_combined_schedule_tuning.yaml
@@ -57,7 +57,7 @@ training:
     - betfred
     - bwin
     markets:
-    - h2h
+    - 1x2
     outcome: home
     closing_tier: closing
     closing_source_priority:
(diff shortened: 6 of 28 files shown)


Acusick1 (commented):
## Review — PR #312

**Approach is solid** — separating 1x2 and h2h at the storage layer is the right fix, and the migration strategy (relabel rows + patch raw_data JSON) is thorough.

### Blocking: MCP server not updated

The `odds_mcp/server.py` has **9+ hardcoded `market="h2h"` calls** (`get_latest_snapshot`, `extract_odds_from_snapshot`). After the migration relabels EPL raw_data JSON keys from `"h2h"` to `"1x2"`, the MCP server will silently return no odds for EPL events — `raw_data_has_market(snapshot.raw_data, "h2h")` will return `False` for any migrated snapshot.

This breaks `get_current_odds`, `get_odds_history`, `get_sharp_soft_spread`, `get_match_brief`, `get_event_features`, and `get_predictions` for EPL. The agent's primary data path goes dark.

The MCP server needs to either:
- Accept a `market` parameter and pass it through, or
- Query for both `"h2h"` and `"1x2"` where moneyline data is needed, or
- Infer the market from the sport/league

### Minor observations

1. **`lineup_signal_check.py`** (experiments script, line 888): still uses `market="h2h"` and doesn't pass `market_key` to `calculate_devigged_bookmaker_target`. Won't affect production but would silently miss EPL data if reused.

2. **Test coverage gap**: `test_sequence_loader.py` 3-way soccer tests (line 618-628) still create odds with `market_key="h2h"` (the `_make_odds` default) and call `extract_devigged_h2h_probs` without passing `market_key`. These tests pass but don't exercise the actual production path where soccer odds have `market_key="1x2"`.

3. **Migration**: the upgrade SQL is complex (nested `jsonb_set` + `jsonb_agg` + subqueries) but correct. The downgrade reverses cleanly. On a large table the correlated subquery in the `WHERE` clause could be slow — consider whether an `EXISTS` subquery or a CTE would perform better for the 82K row update.

4. **Agent prompt** (`agents/mlb/base.md` line 23, 99): still says `market="h2h"` for MLB paper bets. MLB 2-way bets should indeed use `h2h`, so this is correct — just flagging for awareness since someone might confuse it with the 1x2 change.

Acusick1 (commented):
## Re-review — PR #312

The second commit addresses all three issues from the first review:

1. **MCP server** — all 9 hardcoded `market="h2h"` calls replaced with an explicit `market: MarketKey` parameter (`Literal["h2h", "1x2", "totals", "spreads"]`). Clean approach — the agent must now declare which market it wants, which is correct since the MCP server is sport-agnostic.
2. **`lineup_signal_check.py`** — updated to `market="1x2"` + `market_key="1x2"`.
3. **Test fixtures** — 3-way soccer tests in `test_sequence_loader.py` now create odds with `market="1x2"` and pass `market_key="1x2"` to the functions under test. The `_make_odds` helper in `TestCalculateDeviggedBookmakerTarget` also gained the `market` parameter. Good cleanup.

### One remaining item (non-blocking)

**Agent prompts need `market=` guidance.** The MCP tools now *require* `market` as a parameter, but the EPL agent prompts (`agents/epl/checkpoint1.md`, `base.md`) don't mention what value to pass. The EPL agent calls `save_match_brief`, `get_current_odds`, `get_sharp_soft_spread`, etc. — it needs to know to use `market="1x2"`. The MLB prompt mentions `market="h2h"` for `paper_bet` but not for the tools that gained the new required param (`get_sharp_soft_spread`, `get_event_features`, `save_match_brief`, `get_odds_history`).

The LLM agent will likely infer the right market from the tool description + context, but explicit guidance in the prompts would reduce wasted tool calls. Fine to defer to a follow-up.

### CI

Checks still pending — verify they pass before merging.

**Looks good to merge** once CI is green.

PR #317, opened 2026-04-15
Fix compound name resolution in local scheduler backend

### LINKED ISSUE #313: Fix compound name resolution in local scheduler backend
## Goal

`LocalSchedulerBackend.schedule_next_execution()` passes compound job names (e.g., `fetch-oddsportal-epl`) directly to `get_job_function()`, which only recognizes base names from `_JOB_MODULE_MAP`. This is a latent bug that will cause `KeyError` for any per-sport job on the local backend. It hasn't manifested because the local scheduler hasn't been used with per-sport OddsPortal/scores jobs in practice, but it blocks adding new per-sport jobs (agent scheduling).

## Scope

- Resolve compound name via `resolve_job_name()` before calling `get_job_function()` in `schedule_next_execution`
- Remove the redundant second `resolve_job_name()` call that currently only extracts the sport

## Location

- `packages/odds-lambda/odds_lambda/scheduling/backends/local.py` — `schedule_next_execution()`, lines 222-234

## Success criteria

- [ ] `schedule_next_execution("fetch-oddsportal-epl", ...)` resolves to the `fetch-oddsportal` job function with `sport="soccer_epl"` in the context
- [ ] Existing non-compound jobs (`check-health`, `update-status`) still work unchanged
- [ ] Test covers compound name scheduling on local backend

Acusick1:
## Summary
- Resolve compound job names (e.g. `fetch-oddsportal-epl`) via `resolve_job_name()` **before** calling `get_job_function()` in `schedule_next_execution()`, which only recognizes base names from `_JOB_MODULE_MAP`
- Remove the redundant second `resolve_job_name()` call that previously only extracted the sport
- Add tests for both compound and non-compound job name scheduling

## Closes #313

### DIFF
--- a/packages/odds-lambda/odds_lambda/scheduling/backends/local.py
+++ b/packages/odds-lambda/odds_lambda/scheduling/backends/local.py
@@ -222,10 +222,9 @@ class LocalSchedulerBackend(SchedulerBackend):
             # Get job function from centralized registry
             from odds_lambda.scheduling.jobs import JobContext, get_job_function, resolve_job_name
 
-            job_func = get_job_function(job_name)
-
-            # Build JobContext from compound name + payload
-            _, resolved_sport = resolve_job_name(job_name)
+            # Resolve compound name (e.g. "fetch-oddsportal-epl" -> "fetch-oddsportal" + "soccer_epl")
+            base_name, resolved_sport = resolve_job_name(job_name)
+            job_func = get_job_function(base_name)
             ctx_payload: dict[str, object] = {}
             if resolved_sport:
                 ctx_payload["sport"] = resolved_sport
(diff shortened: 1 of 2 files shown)


Acusick1 (commented):
LGTM. Clean fix — correctly resolves compound names before the registry lookup, and the redundant second `resolve_job_name()` call is removed.

**Tests**: Good that real `resolve_job_name` runs (not mocked) so actual parsing logic is exercised. Both compound and non-compound paths covered with context capture.

**Minor nit** (non-blocking): `asyncio.wait_for(asyncio.sleep(0.5), timeout=2.0)` is equivalent to just `await asyncio.sleep(0.5)` since the sleep always finishes before the timeout. Consistent with the existing test at line 252, but the `wait_for` wrapper adds nothing.

CI green. Cannot approve (Claude co-authored).

Issue #315, opened 2026-04-15
Create agent-run job module with self-scheduling


Acusick1:
## Goal

Add an agent job that self-schedules through the existing scheduler, waking the Claude Code agent at appropriate times relative to fixture kickoffs. The agent decides what to do on each wake-up (research, update briefs, place bets, settle) — the scheduler just controls timing. The agent can override the default wake-up interval via MCP tool (#319).

## Scope

- Create `packages/odds-lambda/odds_lambda/jobs/agent_run.py` with the standard `async def main(ctx: JobContext)` entry point
- Implement fixture-proximity scheduling (query DB for next matchday, compute wake interval)
- Implement `claude -p` subprocess invocation
- Self-schedule the next wake-up after each run
- After subprocess exits, check `agent_wakeups` table for agent-requested wake-up time. If present, use it instead of the default tier interval.

## Implementation approach

Follow the `fetch_oddsportal.py` pattern: pre-schedule next run before doing work (survives crashes), then execute, then optionally reschedule at updated interval.

**The agent owns its workflow, not the scheduler.** Each wake-up runs the `/agent` skill with no checkpoint argument — the agent reads fixture schedule and existing briefs from the DB, decides what action is needed (research, bet, settle, skip), and executes. Match briefs in the DB provide cross-session memory.

**Key functions:**
- `_get_next_kickoff(sport_key) -> datetime | None` — earliest scheduled event
- `_compute_wake_interval(hours_until_ko) -> float` — fixture-proximity tiers for agent wake-ups
- `_run_claude_agent(sport) -> int` — spawn `claude -p "/agent {sport}"` via `asyncio.create_subprocess_exec` with timeout and stdout logging
- `_check_agent_requested_wakeup(sport_key) -> datetime | None` — read from `agent_wakeups` table for agent-requested override
- `main(ctx)` — query DB for next fixture, pre-schedule next wake-up at default interval, run claude, check for agent-requested wake-up override, reschedule if needed

**Wake interval tiers (suggested starting points, tune after real use):**

| Hours until next KO | Wake interval | Rationale |
|---|---|---|
| > 48h | 24h | Far out, just check for new fixtures |
| 24-48h | 12h | Research window opening |
| 6-24h | 4h | Active research, early bets possible |
| 1.5-6h | 1h | Lineups dropping, decision time |
| < 1.5h | No wake | Too close to KO, previous run should have decided |
| No fixtures in 7d | 12h | Off-season / international break check-in |

[...]

PR #321, opened 2026-04-15
Add MCP tools for agent scheduling (wakeup + job visibility)

### LINKED ISSUE #319: Add MCP tools for agent scheduling (wakeup + job visibility)
## Goal

Give the agent visibility into the scheduler and control over its own wake-up timing. The agent can see what jobs are scheduled (including its own next wake-up and scrape times) and request an earlier wake-up when it finds something worth checking back on sooner.

## Scope

**DB model:**
- Create `agent_wakeups` table: `id`, `sport_key`, `requested_time` (datetime), `reason` (text), `created_at`. One active row per sport (upsert on sport_key).

**MCP tools (in `odds_mcp/server.py`):**
- `get_scheduled_jobs(sport?)` — list all currently scheduled jobs from the scheduler backend. Wraps `backend.get_scheduled_jobs()` with optional sport filter. Returns job name, next run time, status.
- `schedule_next_wakeup(sport, delay_hours, reason)` — request the agent's next wake-up at `now + delay_hours`. Writes/upserts to `agent_wakeups` table. The agent job module (#315) reads this after subprocess exits and reschedules if the requested time is sooner than the default.

**Shared backend access:**
- Both MCP tools and CLI use `get_scheduler_backend()` from `odds_lambda.scheduling.backends`. No duplication of scheduling logic.

## Location

- `packages/odds-core/odds_core/` — new model for `agent_wakeups`
- `packages/odds-mcp/odds_mcp/server.py` — new MCP tool functions
- Alembic migration for new table

## Implementation approach

`get_scheduled_jobs` is a direct read from the scheduler backend — same code path as `odds scheduler list-jobs` CLI command. For the local APScheduler backend, this requires the scheduler to be running in a separate process; if it's not running, return an informative message rather than an error.

`schedule_next_wakeup` writes to the DB rather than directly calling the scheduler backend, because the MCP server and the scheduler run in separate processes (can't share APScheduler in-memory state). The `agent_run.py` job module (#315) reads from this table after the agent subprocess exits and reschedules accordingly.

## Constraints

- `schedule_next_wakeup` must be idempotent per sport — upsert, not insert. Only the latest request matters.
- `get_scheduled_jobs` must handle the case where the scheduler backend is not running gracefully.

## Notes

- The `agent_wakeups` table is a communication channel between the MCP server process and the scheduler process. It's intentionally simple — one active row per sport, consumed after use.
- The agent might call `get_scheduled_jobs` to see when the next scrape is scheduled, which helps it decide whether to request a scrape refresh or wait.

## Success criteria

- [ ] Agent can call `get_scheduled_jobs()` and see all scheduled jobs with next run times
- [ ] Agent can call `schedule_next_wakeup("epl", 2.0, "lineup news developing")` and the wakeup is stored
- [ ] `agent_run.py` (#315) can read and consume the stored wakeup after subprocess exits
- [ ] Multiple calls to `schedule_next_wakeup` for the same sport upsert (latest wins)

Depends on #314. Consumed by #315.

Acusick1:
## Summary
- Add `AgentWakeup` SQLModel and `agent_wakeups` table (one active row per sport, upsert semantics) with `consumed_at` field for consumption tracking
- Add `get_scheduled_jobs(sport?)` MCP tool — reads from scheduler backend with optional sport filter, graceful handling when backend is unavailable
- Add `schedule_next_wakeup(sport, delay_hours, reason)` MCP tool — upserts to `agent_wakeups` table so `agent_run.py` (#315) can read and reschedule after subprocess exits
- Add Alembic migration for new table
- Add 9 unit tests covering upsert behavior, delay clamping, backend unavailability, and sport filtering

## Closes #319

### DIFF
--- a/migrations/env.py
+++ b/migrations/env.py
@@ -4,6 +4,7 @@ import asyncio
 from logging.config import fileConfig
 
 from alembic import context
+from odds_core.agent_wakeup_models import AgentWakeup  # noqa: F401
 from odds_core.config import get_settings
 from odds_core.epl_data_models import EspnFixture, EspnLineup, FplAvailability  # noqa: F401
 from odds_core.game_log_models import NbaTeamGameLog  # noqa: F401
--- a/migrations/versions/a7b3c9d2e1f0_add_agent_wakeups_table.py
+++ b/migrations/versions/a7b3c9d2e1f0_add_agent_wakeups_table.py
@@ -1,0 +1,41 @@
+"""add agent_wakeups table
+
+Revision ID: a7b3c9d2e1f0
+Revises: 920aee156d9c
+Create Date: 2026-04-15 18:00:00.000000
+
+"""
+
+import sqlalchemy as sa
+import sqlmodel
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "a7b3c9d2e1f0"
+down_revision = "920aee156d9c"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "agent_wakeups",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("sport_key", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.Column("requested_time", sa.DateTime(timezone=True), nullable=False),
+        sa.Column("reason", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.Column("created_at", sa.DateTime(timezone=True), nullable=False),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("sport_key", name="uq_agent_wakeup_sport_key"),
+    )
+    op.create_index(
+        op.f("ix_agent_wakeups_sport_key"),
+        "agent_wakeups",
+        ["sport_key"],
+        unique=False,
+    )
+
+
+def downgrade() -> None:
+    op.drop_index(op.f("ix_agent_wakeups_sport_key"), table_name="agent_wakeups")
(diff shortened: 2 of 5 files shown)


Acusick1 (commented):
## Review — MCP tools for agent scheduling

Clean implementation. Model, tools, migration, and tests are well-structured. One bug and one minor note:

### Bug: upsert doesn't reset `consumed_at`

`schedule_next_wakeup` omits `consumed_at` from `on_conflict_do_update.set_`, so when a wakeup that was already consumed gets upserted with a new request, the old `consumed_at` timestamp persists. The agent_run consumer (#315) will presumably skip rows where `consumed_at IS NOT NULL`, making the new request invisible.

Fix: add `"consumed_at": None` to `set_` in the upsert:

```python
.on_conflict_do_update(
    constraint="uq_agent_wakeup_sport_key",
    set_={
        "requested_time": requested_time,
        "reason": reason,
        "created_at": now,
        "consumed_at": None,  # reset — this is a new request
    },
)
```

A test case for this scenario (insert → mark consumed → upsert → verify consumed_at is None) would be worth adding.

### Minor: `get_scheduler_backend()` is called per-invocation

`get_scheduler_backend()` reads settings and instantiates a backend object on every tool call. For the local backend this connects to APScheduler's TCP socket. Not a problem at current call volumes — just noting it. If it ever matters, a module-level cached instance would work.

---

Everything else looks good: timezone handling is correct (`datetime.now(UTC)`), clamping range is reasonable, the `BackendUnavailableError` graceful handling matches the issue spec, upsert semantics are right (minus the consumed_at bug above), tests cover the key scenarios including coexistence of different sports. CI passes.

PR #323, opened 2026-04-15
Simplify agent scheduling: agent-driven wake times + bootstrap

### LINKED ISSUE #316: Bootstrap agent jobs in scheduler start command
## Goal

When `odds scheduler start` runs, it should bootstrap agent scheduling alongside existing data pipeline jobs. This kicks off the self-scheduling chain so the agent wakes up automatically based on fixture proximity.

## Scope

- Update `start_local()` in the scheduler CLI to call `agent_run.main(JobContext(sport=...))` for each configured sport after existing bootstrap
- The bootstrap call computes the next wake time from the DB and schedules it — no claude invocation

## Location

- `packages/odds-cli/odds_cli/commands/scheduler.py` — `start_local()` / `run_scheduler()`

## Implementation approach

After the existing `fetch_odds.main()` bootstrap block, iterate configured sports and bootstrap agent scheduling:

```python
for sport_key, suffix in [("soccer_epl", "epl"), ("baseball_mlb", "mlb")]:
    try:
        await agent_main(JobContext(sport=sport_key))
        console.print(f"[green]  agent-run-{suffix} bootstrapped[/green]")
    except Exception as e:
        console.print(f"[yellow]  agent-run-{suffix} bootstrap failed: {e}[/yellow]")
```

Agent bootstrap failure should not block the scheduler from starting.

## Success criteria

- [ ] `odds scheduler start` shows agent bootstrap status for each sport
- [ ] `odds scheduler list-jobs` shows scheduled agent wake-ups after bootstrap
- [ ] Agent bootstrap failure does not prevent other jobs from running

Depends on #315.

Acusick1:
## Summary
- **Agent decides its own schedule**: removed proximity tier computation from Python code (`_compute_wake_interval`, `_should_skip_run`, `schedule_next()`, all tier constants). The agent uses the `schedule_next_wakeup` MCP tool to write its desired next wake time to the `agent_wakeups` table.
- **Simpler `main()` flow**: pre-schedule a 12h fallback (crash protection) → run agent → read `agent_wakeups` → override fallback with agent's requested time (overnight clamped). If agent crashes or forgets to schedule, the fallback fires.
- **Bootstrap in `scheduler start`**: schedules an immediate agent run for each configured sport. The agent then self-manages from there.
- **Prompt guidance**: added "Scheduling Your Next Wake-Up" section to both EPL and MLB agent prompts with proximity heuristics.
- Net -151 lines — simpler code, smarter agent.

## Closes #316

### DIFF
--- a/packages/odds-cli/odds_cli/commands/scheduler.py
+++ b/packages/odds-cli/odds_cli/commands/scheduler.py
@@ -58,9 +58,10 @@ def start_local():
         # Bootstrap by running initial fetch to start self-scheduling
         console.print("[green]Running initial fetch to bootstrap scheduler...[/green]")
 
+        from odds_lambda.scheduling.jobs import JobContext
+
         try:
             from odds_lambda.jobs import fetch_odds
-            from odds_lambda.scheduling.jobs import JobContext
 
             await fetch_odds.main(JobContext())
             console.print("[green]  fetch-odds bootstrapped[/green]")
@@ -73,9 +74,10 @@ def start_local():
         # Bootstrap agent jobs for each configured sport
         from odds_lambda.jobs import agent_run
 
-        for sport_key, suffix in [("soccer_epl", "epl"), ("baseball_mlb", "mlb")]:
+        for sport_key in app_settings.data_collection.sports:
+            suffix = sport_key.split("_")[-1]
             try:
-                await agent_run.main(JobContext(sport=sport_key))
+                await agent_run.schedule_next(sport_key)
                 console.print(f"[green]  agent-run-{suffix} bootstrapped[/green]")
             except Exception as e:
                 console.print(f"[yellow]  agent-run-{suffix} bootstrap failed: {e}[/yellow]")
--- a/packages/odds-cli/odds_cli/commands/scheduler.py
+++ b/packages/odds-cli/odds_cli/commands/scheduler.py
@@ -63,12 +63,24 @@ def start_local():
             from odds_lambda.scheduling.jobs import JobContext
 
             await fetch_odds.main(JobContext())
-            console.print("[green]✓ Bootstrap complete[/green]\n")
+            console.print("[green]  fetch-odds bootstrapped[/green]")
         except Exception as e:
-            console.print(f"[bold red]✗ Bootstrap failed:[/bold red] {e}\n")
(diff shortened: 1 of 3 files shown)


Acusick1 (commented):
## Review

Clean extraction — `schedule_next()` is the right seam for bootstrap-without-subprocess. Two issues worth addressing:

### 1. Missing sport validation in `schedule_next()`

`main()` has an explicit guard:
```python
if sport not in OVERNIGHT_WINDOWS:
    raise ValueError(f"No overnight window configured for {sport} ...")
```

`schedule_next()` does a bare `OVERNIGHT_WINDOWS[sport]` dict access (line ~167 in the new code). When called from bootstrap with a sport that has no overnight window configured, this raises an unhelpful `KeyError` instead of the descriptive `ValueError`. Either move the guard into `schedule_next()` or add one there too — the bootstrap caller in `scheduler.py` catches `Exception` so it won't crash, but the error message shown to the user will be cryptic.

### 2. No tests for `schedule_next()` or the bootstrap path

Existing tests cover `main()` orchestration (which now delegates to `schedule_next()`), so they implicitly test the scheduling logic still works. But there's no test for:
- `schedule_next()` called standalone (the bootstrap use case)
- The scheduler bootstrap loop in `scheduler.py`

At minimum a unit test for `schedule_next()` directly (mock `get_next_kickoff` + `self_schedule`, assert it returns `(next_time, hours_until_ko)`) would catch regressions in the extracted function independently of `main()`.

### Minor nits

- `compound_job_name` and overnight window resolution are now computed in both `schedule_next()` and `main()`. This is necessary since `main()` needs them for the override section, but worth a brief comment in `main()` noting why (or consider having `schedule_next()` return them to avoid the duplication).
- `schedule_next()` also calls `get_settings()` redundantly (already called in `main()`). Negligible cost since it's likely cached, just noting it.

### What looks good

- Bootstrap failures isolated per-sport — won't block the scheduler
- Sports derived from config, not hardcoded
- `sport` added to `agent_proximity` log event — helpful for multi-sport debugging
- CI passes

Acusick1 (commented):
All three items addressed in c56419c — looks good.

- `schedule_next()` now owns the `ValueError` guard for unknown sports
- `ScheduleResult` NamedTuple eliminates the duplicated computation in `main()` — clean
- Three standalone tests cover the bootstrap use case

CI green. LGTM (commenting only — co-authored by Claude).

PR #328, opened 2026-04-16
Upgrade to APScheduler 4 and merge scrape worker into scheduler

### LINKED ISSUE #326: Upgrade to APScheduler 4 and merge scrape worker into scheduler
## Goal

The scrape worker (`odds worker start`) and scheduler (`odds scheduler start`) are separate processes on the same machine. The worker processes on-demand scrape jobs enqueued by the agent via MCP (`refresh_scrape`), using a `ScrapeJob` DB table as a cross-process queue. This exists because APScheduler 3.x job stores cannot be shared between processes.

APScheduler 4.x redesigns this: a shared `SQLAlchemyDataStore` + `AsyncpgEventBroker` (both backed by the existing Postgres, no new infrastructure) lets any process add jobs that the scheduler picks up immediately via LISTEN/NOTIFY. This eliminates the worker process, the `ScrapeJob` queue table, and the polling loop — the MCP server calls `add_job()` directly and the scheduler executes it.

## Scope

- Upgrade `apscheduler` from 3.11.x to 4.0.0a6 (pin exact version)
- Rewrite `LocalSchedulerBackend` to use `AsyncScheduler` + `SQLAlchemyDataStore` + `AsyncpgEventBroker`
- Replace MCP `refresh_scrape` — instead of inserting a `ScrapeJob` row, call `scheduler.add_job()` on a shared `AsyncScheduler` instance backed by the same Postgres data store
- Add `asyncio.Semaphore(1)` in `run_harvester_upcoming()` to serialize Playwright access
- Remove `odds worker start` CLI command
- Drop `ScrapeJob` model and `scrape_jobs` table (alembic migration)
- Update or remove `get_scrape_status` MCP tool

## Location

- `packages/odds-core/pyproject.toml` — upgrade apscheduler dependency
- `packages/odds-lambda/odds_lambda/scheduling/backends/local.py` — rewrite to APScheduler 4 API
- `packages/odds-mcp/odds_mcp/server.py` — `refresh_scrape` and `get_scrape_status` tools
- `packages/odds-cli/odds_cli/commands/scheduler.py` — update bootstrap to APScheduler 4 context manager
- `packages/odds-cli/odds_cli/commands/worker.py` — delete
- `packages/odds-core/odds_core/scrape_job_models.py` — delete
- New: `packages/odds-lambda/odds_lambda/browser_lock.py` — Playwright semaphore
- `packages/odds-lambda/odds_lambda/jobs/fetch_oddsportal.py` — acquire semaphore

## Implementation approach

**APScheduler 4 setup** (in `LocalSchedulerBackend` or replacing it):

```python
from apscheduler import AsyncScheduler
from apscheduler.datastores.sqlalchemy import SQLAlchemyDataStore
from apscheduler.eventbrokers.asyncpg import AsyncpgEventBroker

engine = create_async_engine(database_url)
data_store = SQLAlchemyDataStore(engine)
event_broker = AsyncpgEventBroker.from_async_sqla_engine(engine)

async with AsyncScheduler(data_store, event_broker) as scheduler:
    await scheduler.run_until_stopped()
```

**Self-scheduling pattern**: keep the existing data-driven proximity scheduling. Jobs compute next run time from game proximity and call `add_job()` with a `DateTrigger`. APScheduler 4's built-in triggers (Interval, Cron) can't express this dynamic logic, so the self-scheduling pattern stays:

```python
await scheduler.add_job(main, trigger=DateTrigger(run_date=next_time), args=[ctx])
```

**MCP `refresh_scrape`**: create an `AsyncScheduler` with the same data store and event broker (separate process, shared DB). `add_job()` without a trigger runs immediately; the event broker notifies the scheduler via LISTEN/NOTIFY:

```python
async with AsyncScheduler(data_store, event_broker) as scheduler:
    await scheduler.add_job(ingest_league, args=[spec])
```

**Playwright semaphore**: module-level, acquired inside `run_harvester_upcoming()`:

```python
# browser_lock.py
import asyncio
playwright_semaphore = asyncio.Semaphore(1)
```

**AWS backend**: unaffected. It uses EventBridge, not APScheduler. The `SchedulerBackend` interface (`schedule_next_execution`) is our abstraction — only the local backend's internals change.

## Notes

- `asyncpg` is already a dependency in `odds-core`.
- APScheduler 4 creates its own tables on startup (`SQLAlchemyDataStore` handles schema).
- The `separate_worker` [example](https://github.com/agronholm/apscheduler/tree/master/examples/separate_worker) in the APScheduler repo is almost exactly our scheduler/MCP architecture.
- 4.0.0a6 is pre-release. Acceptable risk for a paper trading pipeline. Pin the exact version.

## Success criteria

- [ ] `apscheduler==4.0.0a6` in dependencies
- [ ] `odds scheduler start` uses `AsyncScheduler` with Postgres-backed data store and event broker
- [ ] MCP `refresh_scrape` triggers a scrape job that the scheduler picks up without polling
- [ ] Only one Playwright instance runs at a time
- [ ] `odds worker start` removed
- [ ] `scrape_jobs` table dropped
- [ ] Existing self-scheduling jobs (`fetch-oddsportal`, `agent-run`) continue to work

Depends on #325.

Acusick1:
## Summary

- **Upgrade APScheduler 3.x → 4.0.0a6** with `AsyncScheduler` + `SQLAlchemyDataStore` + `AsyncpgEventBroker` (Postgres LISTEN/NOTIFY for cross-process job coordination)
- **Rewrite `LocalSchedulerBackend`** to use APScheduler 4 API — self-scheduling pattern preserved via `DateTrigger` + `add_schedule` with `ConflictPolicy.replace`
- **MCP `refresh_scrape` now adds jobs directly** via a temporary `AsyncScheduler` sharing the same Postgres data store — LISTEN/NOTIFY notifies the running scheduler immediately, eliminating the polling-based `ScrapeJob` queue
- **Add Playwright semaphore** (`browser_lock.py`) to serialize headless browser access
- **Remove `odds worker start`** CLI command — no longer needed
- **Drop `scrape_jobs` table** and `ScrapeJobStatus` enum via alembic migration
- **Remove `odds scrape job-status`** CLI subcommand (queried the now-deleted table)
- **Add deduplication guard** in `refresh_scrape` to prevent duplicate scrape jobs for the same league/market
- **Extract `build_scheduler()`** as shared helper used by both `LocalSchedulerBackend` and MCP tools

## Closes #326


Acusick1 (commented):
## Review — APScheduler 4 migration

CI green, issue scope covered, clean deletion of the worker/ScrapeJob infrastructure. A few items to consider:

### 1. `get_scrape_status` observability regression

The old tool returned completion stats (`matches_scraped`, `events_matched`, etc.) and error messages. The new version only lists pending jobs in the APScheduler data store — once a job completes or fails, it vanishes from the queue with no way to check results. This is a meaningful downgrade for the agent workflow (submit scrape → poll until done → see stats). Worth either:
- Logging completion stats to structlog so they're at least visible in logs, or
- Tracking that this is a known gap to address later (e.g., APScheduler 4 job events/listeners could publish results)

### 2. `get_job_status` bare `except Exception` (local.py:158)

```python
try:
    sched = await self._scheduler.get_schedule(job_name)
except Exception:
    return None
```

This swallows DB connection errors, serialization bugs, etc. Should catch `ScheduleLookupError` specifically (like `cancel_scheduled_execution` does on line 245).

### 3. `refresh_scrape` dedup via `j.args[0]` inspection (server.py:~1013–1021)

The dedup check introspects job args with `hasattr(existing_spec, "league")`. This is brittle — if `ingest_league`'s signature changes or args are serialized differently by the data store, dedup silently stops working with no error. Consider whether a simpler approach (e.g., deterministic job ID like `scrape-{league}-{market}`) would be more robust.

### 4. Minor: `refresh_scrape` creates engine per call

Each MCP call spins up a full `build_scheduler()` → `create_async_engine` → `SQLAlchemyDataStore` → `AsyncpgEventBroker`. For an infrequent tool this is fine, but if call volume increases it'll be worth caching or pooling. Not blocking.

### 5. Alpha pin is acknowledged

`apscheduler==4.0.0a6` — issue notes this explicitly, reasonable for paper trading pipeline. Just flagging for visibility.

---

Overall this is a clean simplification — removes an entire process, a DB table, and a polling loop in favor of LISTEN/NOTIFY. The items above are mostly hardening for when things go wrong, not blockers.

Acusick1 (commented):
## Updated review (after running tests and reading source files)

My earlier review was based on the diff only. After reading the actual files, running the full test suite (141 passed, 1 local-only migration failure from unapplied `alembic upgrade head`), and verifying the APScheduler 4 API, here's the corrected assessment.

### Confirmed: clean implementation

- No stale references to `ScrapeJob`, `ScrapeJobStatus`, `scrape_job_models`, or `worker` anywhere in the codebase
- APScheduler 4 API usage is correct: MCP tools use `add_job()`/`get_jobs()` for one-shot executions, `LocalSchedulerBackend` uses `add_schedule()`/`get_schedules()` for self-scheduling with `DateTrigger` — these are distinct concepts in v4
- `Job` objects have `task_id`, `args`, `created_at` — all attributes used in dedup and status reporting
- All 23 scheduler integration tests pass (including execution, compound names, and E2E)
- Migration is correct (drop indexes → drop table → drop enum, with reversible downgrade)

### Remaining findings (minor, non-blocking)

**1. `get_job_status` bare `except Exception` (local.py:158)**
Should catch `ScheduleLookupError` specifically (like `cancel_scheduled_execution` does on line 245), otherwise DB connection errors are silently swallowed.

**2. `get_scrape_status` loses completed job info**
The old tool returned ingestion stats and error messages. The new version only shows pending jobs — once APScheduler runs a job, it's gone from `get_jobs()`. Not blocking for initial merge but worth noting as a gap for the agent workflow.

**3. Engine-per-call in MCP tools**
Each `refresh_scrape`/`get_scrape_status` call creates a full `build_scheduler()` → engine → data store → event broker. Fine at current volume, worth caching if call frequency increases.

Overall this is a clean simplification that eliminates a process, a table, and a polling loop.

Acusick1 (commented):
## Review — APScheduler 4 migration (latest commit `a48f94a`)

### 🔴 Blocker: CI is failing

4 tests in `tests/unit/test_agent_wakeup.py::TestGetScheduledJobs` still patch `odds_mcp.server.get_scheduler_backend`, but commit `e163798` removed that indirection — `get_scheduled_jobs` now calls `build_scheduler()` directly:

```
AttributeError: module 'odds_mcp.server' does not have the attribute 'get_scheduler_backend'
```

Tests to update (`test_backend_unavailable_returns_message`, `test_sport_filtering`, `test_no_filter_returns_all`, `test_generic_exception_returns_message`) — they need to mock `build_scheduler` (or the returned `AsyncScheduler`) instead. The `test_backend_unavailable_returns_message` assertion `"unavailable" in result["message"].lower()` will also fail since the new error path returns `"Could not query scheduler: {e}"`.

### 🟡 Latent bug: `odds scheduler list-jobs` silently returns empty

`packages/odds-cli/odds_cli/commands/scheduler.py:249-257` still does:

```python
backend = get_scheduler_backend()
jobs = await backend.get_scheduled_jobs()
```

But `LocalSchedulerBackend.get_scheduled_jobs` short-circuits to `[]` when `self._scheduler is None` (local.py:137-139), and this CLI path never enters the backend's `__aenter__`. Result: the command prints "No jobs currently scheduled" every time on the local backend, with only a `local_scheduler_not_started` warning in structlog. The MCP `get_scheduled_jobs` solved this by querying the shared data store via `build_scheduler()` — the CLI should follow the same pattern.

### 🟡 Inefficient per-call scheduler construction

- `self_schedule()` inside the running scheduler process goes through `get_scheduler_backend()` → fresh `LocalSchedulerBackend` with `_scheduler=None` → `_add_schedule` else-branch → spins up an ephemeral `AsyncScheduler` (new data store + new `AsyncpgEventBroker` that sets up LISTEN/NOTIFY) on every self-schedule. Correct but wasteful when we're already inside a running scheduler.
- `refresh_scrape`, `get_scrape_status`, `get_scheduled_jobs` each build a full `AsyncScheduler` stack per MCP call. Low volume for now; worth a shared cached `AsyncScheduler` (or at least a shared engine + lazy data store) if call frequency grows.
- `get_scheduled_jobs` uses `SchedulerRole.scheduler` — a read-only tool briefly taking on scheduler duties on each call. Probably benign, worth noting.

### 🟢 Confirmed clean

- No stale `ScrapeJob`/`ScrapeJobStatus`/`scrape_job_models`/`worker` references anywhere
- Migration `917adbe0576d_drop_scrape_jobs_table.py` is reversible (index → table → enum, clean downgrade)
- `_add_schedule` dual-mode (context-manager vs ephemeral) is well-factored; `configure_task` caching in `_configured_tasks` is a nice touch
- `browser_lock.py` semaphore is correctly process-local — the scheduler process is now the only executor, so single-process serialization is sufficient
- Agent-requested wakeup override (a48f94a) correctly applies regardless of direction and skips bootstrap when a schedule already exists
- `get_job_status` narrowed to `ScheduleLookupError` (fixed since first round)
- `apscheduler==4.0.0a6` alpha pin acknowledged and acceptable for paper trading

### 📝 Known gap (non-blocking)

`get_scrape_status` no longer surfaces completion stats (`matches_scraped`, `events_matched`) or error messages from completed jobs — once APScheduler runs a job it disappears from `get_jobs()`. The agent workflow (submit scrape → poll until done → see stats) is degraded. Could be restored later via APScheduler 4 job event listeners that log results to structlog or a small results table.

---

Overall the architecture is a clean win — entire worker process, table, and polling loop replaced by LISTEN/NOTIFY. CI fix and the `list-jobs` bug are the main items to address before merge.

Acusick1:
## Correction to inefficiency point

Dug into the APScheduler 4 source (`4.0.0a6` in our venv) and docs. My earlier "inefficiency" bullet was too broad — splitting into what's actually avoidable vs what isn't:

### ✅ You're right: `refresh_scrape` write path is canonical

The `separate_worker` [example](https://github.com/agronholm/apscheduler/blob/master/examples/separate_worker/async_scheduler.py) is precisely `build engine → SQLAlchemyDataStore → AsyncpgEventBroker → async with AsyncScheduler(...) as s: await s.add_schedule(...) → exit` with the inline comment "we don't actually start the scheduler here!". Notifying the running scheduler requires publishing on the asyncpg channel, which requires an `AsyncpgEventBroker`. Per-call construction in `refresh_scrape` is the documented pattern. No action needed.

### 🟡 But: `self_schedule()` inside a running job IS avoidable

APScheduler 4 exposes `current_async_scheduler` / `current_job` as module-level contextvars, set automatically during job execution:

- `apscheduler/_context.py` defines them; re-exported from `apscheduler/__init__.py:5-7`
- `AsyncScheduler.run_until_stopped()` sets the contextvar at `_schedulers/async_.py:871`
- Docs: [userguide.rst §"Context variables"](https://github.com/agronholm/apscheduler/blob/master/docs/userguide.rst) (L523-544)
- [versionhistory.rst:167-168](https://github.com/agronholm/apscheduler/blob/master/docs/versionhistory.rst): \"`current_async_scheduler` will always be available when a scheduler is running in the current context\"

So `helpers.py::self_schedule` could do:

```python
from apscheduler import current_async_scheduler

scheduler = current_async_scheduler.get(None)
if scheduler is not None:
    await scheduler.add_schedule(job_func, DateTrigger(...), id=job_name, ...)
else:
    # fall back to short-lived AsyncScheduler for bootstrap / out-of-process callers
    ...
```

That skips an entire engine + data store + asyncpg LISTEN broker spin-up on every `fetch_oddsportal` / `agent_run` self-reschedule — which runs every time those jobs execute.

### 🟡 And: read-only MCP paths can drop the event broker

`AsyncScheduler.__aenter__` → `_ensure_services_initialized` (`async_.py:177-210`) unconditionally starts both the data store and the event broker. `AsyncpgEventBroker.start()` opens a **dedicated asyncpg connection and registers a LISTEN callback on the `apscheduler` channel** (`eventbrokers/asyncpg.py:102-156`) — real overhead on a pure read.

`SQLAlchemyDataStore.get_schedules()` / `get_jobs()` are public methods (`datastores/sqlalchemy.py:637, 860`) that only touch the DB. For `get_scrape_status` and `get_scheduled_jobs` (both read-only), you can swap `AsyncpgEventBroker` for `LocalEventBroker` (no external connections) and skip the LISTEN setup entirely — still go through `AsyncScheduler` so the lifecycle is correct, but avoid the asyncpg subscription per call. Minor optimisation, not a blocker.

---

Net: the CI failure and `list-jobs` CLI bug from my previous comment are still the blockers. The `self_schedule` contextvar point is a cheap, documented improvement worth considering. The read-path broker swap is optional. The MCP write path is canonical as-is — ignore my earlier "wasteful" framing on `refresh_scrape`.

Acusick1 (commented):
## Re-review of `bd0eb3a`

CI green. All three prior blockers addressed:

- ✅ `TestGetScheduledJobs` now patches `odds_lambda.scheduling.backends.local.build_scheduler` (inline import in `server.py:980` makes source-module patching correct). Assertion for the unavailable case updated to match the real error message.
- ✅ `LocalSchedulerBackend.get_scheduled_jobs` / `get_job_status` fall back to an ephemeral `build_scheduler()` when `_scheduler is None`, fixing `odds scheduler list-jobs`. Read/write paths now symmetric with `_add_schedule`.
- ✅ `_add_schedule` uses `self._scheduler or current_async_scheduler.get(None)`, reusing the running scheduler from inside jobs. `_write_schedule` nicely factors the task-cache + add-schedule logic.

### Minor, non-blocking observations

1. `self._configured_tasks` is shared across `self._scheduler`, `current_async_scheduler`, and ephemeral-scheduler code paths. `configure_task` is idempotent at the APScheduler layer, so no correctness issue — just noting that the cache semantic is now "some scheduler we've used has this task configured", not strictly "self._scheduler has it". Fine.
2. Read-only fallbacks still build full `AsyncScheduler` (with `AsyncpgEventBroker` LISTEN setup) per call. `LocalEventBroker` swap would skip that; noted previously, still not worth addressing here.

### Unresolved from earlier review (flagged as known gap, not blocking)

- `get_scrape_status` no longer surfaces completion stats or error messages — once APScheduler runs a job it's gone from `get_jobs()`. Agent loses the "submit scrape → poll until done → see stats" pattern.

---

LGTM from my end. Ship it.

PR #329, opened 2026-04-17
Per-run agent JSONL trace + dev-mode CLI

Acusick1:
## Summary

Two related pieces of infrastructure for iterating on the EPL/MLB betting agent:

- **Per-run JSONL trace** (`49daeca`): `_run_claude_agent` now spawns `claude -p --output-format stream-json --verbose` and writes the full message stream to `logs/agent_runs/{sport}_<ts>.jsonl`. A Python tee parses each line and surfaces `tool_use` calls and the final `result` summary to the main structlog file as `agent_tool_use` / `agent_run_summary` events — the scheduler log stays readable while the full reasoning trace is preserved per run. Oldest files pruned on write (keep 50 per sport).

- **Dev-mode CLI** (`70c119c`): two entry points for running the agent against a sandbox DB without touching the source-of-truth scheduler:
  - `odds agent run --sport <sport> [--db <name>]` — invokes `_run_claude_agent` directly, bypassing the scheduler wrapper. Defaults to `--db odds_test`.
  - `odds scheduler start` gains `--db` and `--bootstrap/--no-bootstrap`. When `--db` is set, bootstrap defaults to off (sidecar mode — prevents a scheduler-launched `agent-run` from clashing with a manual `odds agent run`). Pass `--bootstrap` explicitly for full-loop smoke tests.
  - `odds_cli.db_override.swap_dbname` is the shared URL helper; `override_database_url` resets the lru-cached `Settings` so subsequent reads see the override.
  - `scripts/snapshot_to_dev.sh` (gitignored — local-only) refreshes the sandbox DB via `pg_dump | pg_restore` inside the postgres docker-compose container.

## Test plan

- [x] Unit tests: `_preview_tool_input`, `_log_stream_message`, `swap_dbname`, `override_database_url` (18 new cases, all green).
- [x] Live end-to-end run against `odds_test`: 9m13s / 64 turns / $3.43. JSONL captured 488 KB / 157 messages. Main log showed `agent_tool_use` for every MCP/tool call with readable multi-kwarg previews and final `agent_run_summary`. 3 briefs landed in `odds_test.match_briefs`; 0 in source `odds` (isolation confirmed).
- [x] CLI smoke tests: `odds scheduler start` with no args (bootstrap on), `--db odds_test` (bootstrap off), `--db odds_test --bootstrap` (bootstrap on).
- [ ] Known limitation — in sandbox mode with no sidecar scheduler, the agent's `refresh_scrape` → `get_scrape_status` poll loop has nothing to pick up the job. Workaround: run `odds scheduler start --db odds_test` in another terminal (bootstrap auto-off means it just listens for submitted jobs).
- [ ] Known limitation — agent filesystem edits (e.g. `packages/odds-mcp/agents/epl/observations.md`) are not sandboxed; only DB writes are isolated. Worth addressing later via git worktree per dev run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,6 +98,9 @@ logs/
 plans/
 *.log
 
+# Local dev helpers
+scripts/snapshot_to_dev.sh
+
 # OS
 .DS_Store
 Thumbs.db
--- a/packages/odds-cli/odds_cli/commands/agent.py
+++ b/packages/odds-cli/odds_cli/commands/agent.py
@@ -1,0 +1,49 @@
+"""Run the sport agent subprocess outside the scheduler wrapper."""
+
+from __future__ import annotations
+
+import asyncio
+
+import typer
+from rich.console import Console
+
+from odds_cli.db_override import override_database_url
+
+app = typer.Typer()
+console = Console()
+
+
+@app.command("run")
+def run(
+    sport: str = typer.Option(
+        ..., "--sport", "-s", help="Sport key (e.g. 'soccer_epl', 'baseball_mlb')"
+    ),
+    db: str = typer.Option(
+        "odds_test",
+        "--db",
+        help="Database name to run against. Swapped into DATABASE_URL.",
+    ),
+) -> None:
+    """Run the sport agent subprocess directly, bypassing the scheduler wrapper.
+
+    No pre-schedule, no agent_wakeups consumption, no reschedule — so repeated
+    runs don't disturb the scheduler. By default the agent runs against the
+    ``odds_test`` database (same host/credentials as the configured
+    DATABASE_URL, just a different dbname); refresh it from the source-of-truth
+    DB with ``scripts/snapshot_to_dev.sh`` before use.
+    """
+    override_database_url(db)
+
+    from odds_lambda.jobs.agent_run import _run_claude_agent
+
+    console.print(f"[bold blue]Running agent for {sport} against db={db}...[/bold blue]")
+    exit_code = asyncio.run(_run_claude_agent(sport))
+
+    if exit_code < 0:
+        console.print("[bold red]Agent timed out[/bold red]")
+        raise typer.Exit(1)
+    if exit_code != 0:
+        console.print(f"[yellow]Agent exited with code {exit_code}[/yellow]")
+        raise typer.Exit(exit_code)
+
+    console.print("[green]Agent finished[/green]")
--- a/packages/odds-cli/odds_cli/commands/scheduler.py
+++ b/packages/odds-cli/odds_cli/commands/scheduler.py
@@ -30,7 +30,7 @@ def start_local(
             "driving manually)."
         ),
     ),
-):
+) -> None:
     """
     Start local scheduler for testing (APScheduler backend).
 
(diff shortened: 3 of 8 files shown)


Acusick1 (commented):
Reviewed against `~/.claude/review-checklist.md`. CI green, scope is tight, and the live end-to-end run + isolation check in the PR description are reassuring. Posting as a comment (not approve) since the PR is co-authored by Claude.

## Observations

**`_run_claude_agent` — StreamReader limit may be too low for stream-json.**
`asyncio.create_subprocess_exec` defaults the stream reader to 64 KiB per line. Under `--output-format stream-json`, every message (including `tool_result` blocks) is a single JSON line, and realistic messages can blow past 64 KiB — e.g. a Read of a large file, a web fetch/search result, or a tool_use with a big `input`/`file_text`. When a line exceeds the limit `readline()` raises `LimitOverrunError`, which would abort the entire agent run. The 488 KB / 157-message live run averages ~3 KB per line so likely didn't trigger it, but it's a latent failure waiting on the first big tool_result.

Suggest bumping the limit when spawning the subprocess, e.g.:
```python
proc = await asyncio.create_subprocess_exec(
    *cmd,
    stdout=asyncio.subprocess.PIPE,
    stderr=asyncio.subprocess.STDOUT,
    cwd=str(settings.project_root),
    limit=8 * 1024 * 1024,  # 8 MiB per line — stream-json tool_results can be large
)
```

**`start_local` missing return type annotation.**
Project CLAUDE.md mandates type hints on every function. `def start_local(db: ..., bootstrap: ...):` in `packages/odds-cli/odds_cli/commands/scheduler.py:17` should be `-> None`. Pre-existing in the original, but the signature was rewritten here so worth fixing in passing.

**Timestamp collision on filename.**
`logs/agent_runs/{sport}_{YYYYMMDDTHHMMSSZ}.jsonl` has 1s resolution. If two `odds agent run` invocations start in the same second they share a filename; `open("ab")` means they'd interleave JSONL lines into the same file. Rare in practice (scheduler only triggers one) but worth a microsecond or a PID suffix to make manual parallel runs safe.

**`_preview_tool_input` doesn't collapse newlines.**
Values with embedded `\n` (e.g. `file_text` for Write, multi-line `command` for Bash) will render as multi-line preview strings in the main log. Minor readability issue; `s.replace(\"\\n\", \" \")` before the truncation fixes it.

## Nits / low priority

- `scripts/snapshot_to_dev.sh` is gitignored. Fine, but a committed template (e.g. `scripts/snapshot_to_dev.sh.example`) would save every dev from reinventing the `pg_dump | pg_restore` invocation.
- The `--db` flag takes the dbname unvalidated and slots it into a URL path. Local-only dev tool, so not a security concern, but `swap_dbname(\"foo?x=1\", ...)` produces a URL with an accidental query string — worth a one-line sanity check (`if \"/\" in new_db or \"?\" in new_db: raise`).

## Positives

- Tee split (full trace to per-run JSONL, tool calls + final summary to main log) is exactly the right shape for iterating on agent prompts without drowning the scheduler log.
- `_log_stream_message` is nicely defensive (non-dict `message`, unknown types, missing fields all handled) and the unit tests cover each branch.
- `override_database_url` + `reset_settings_cache` is the cleanest way to do the env swap — good reuse of the existing cache-reset hook.
- Default-off bootstrap when `--db` is set is a smart default for the sidecar workflow; the override path (`--bootstrap`) is there for full-loop smoke tests.

Acusick1 (commented):
Fixes look good — all five review items addressed in cda567d, CI green. LGTM (posting as a comment since Claude is co-author, but no further blockers from my side).

PR #332, opened 2026-04-18
Harden MCP agent tools: list-param workaround, retail coverage, SportKey

Acusick1:
## Summary
- Coerce stringified array params in MCP tool inputs (workaround for anthropics/claude-code#22394 — Claude Code's MCP client serialises arrays as JSON strings). Helper accepts JSON-array strings and comma-separated strings; remove once upstream is fixed.
- `get_sharp_soft_spread` now returns every non-sharp book in the snapshot with per-book `market_hold` (computed via existing `calculate_market_hold`), dropping the arbitrary 5-book default. `retail_bookmakers` param removed from both `get_sharp_soft_spread` and `get_event_features`.
- New `odds_core.sports.SportKey` literal applied to four MCP tool signatures (`get_upcoming_fixtures`, `get_slate_briefs`, `get_scheduled_jobs`, `schedule_next_wakeup`) so typos like `sport='mlb'` fail validation instead of silently scheduling nothing.
- Agent parallelism guidance updated to dispatch the new user-level `web-research` subagent (which has `WebSearch`/`WebFetch`) instead of `general-purpose`.

## Test plan
- [x] `uv run pytest tests/unit/test_mcp_server.py` — 33 passed (includes new `TestCoerceBookmakerList`)
- [x] `uv run ruff check` / `uv run ruff format` clean
- [x] Reconnect `/mcp` in Claude Code and verify `get_sharp_soft_spread` / `get_event_features` accept list inputs end-to-end — both return full per-book `market_hold` output on Chelsea vs Man Utd 1x2
- [x] Verify an invalid sport key (e.g. `sport='mlb'`) is now rejected by the MCP validation layer — returns `literal_error: Input should be 'soccer_epl' or 'baseball_mlb'`

### DIFF
--- a/packages/odds-analytics/odds_analytics/utils.py
+++ b/packages/odds-analytics/odds_analytics/utils.py
@@ -7,7 +7,7 @@ devigging, h2h winner, profit-from-odds) lives in ``odds_core.odds_math``.
 from __future__ import annotations
 
 import math
-from collections.abc import Sequence
+from collections.abc import Collection, Iterable, Sequence
 from typing import TYPE_CHECKING
 
 from odds_core.odds_math import american_to_decimal, calculate_implied_probability
@@ -135,6 +135,27 @@ def calculate_market_hold(odds_list: list[int]) -> float:
     """Market hold (bookmaker's edge/vig) from a list of odds for all outcomes."""
     implied_probs = [calculate_implied_probability(odds) for odds in odds_list]
     return sum(implied_probs) - 1.0
+
+
+def per_book_market_holds(
+    prices: Iterable[tuple[str, str, int]],
+    required_outcomes: Collection[str],
+) -> dict[str, float]:
+    """Market hold per bookmaker from ``(bookmaker, outcome, price)`` tuples.
+
+    Books missing any of ``required_outcomes`` are dropped — a partial-market
+    sum would understate the hold (possibly negative) with no signal to the
+    caller. Duplicate ``(bookmaker, outcome)`` entries keep the last price.
+    """
+    required = set(required_outcomes)
+    by_book: dict[str, dict[str, int]] = {}
+    for bookmaker, outcome, price in prices:
+        by_book.setdefault(bookmaker, {})[outcome] = price
+    return {
+        bm: calculate_market_hold(list(outcome_prices.values()))
+        for bm, outcome_prices in by_book.items()
+        if set(outcome_prices) >= required
+    }
 
 
 def detect_arbitrage(odds_list: list[tuple[str, int]]) -> tuple[bool, float, dict]:
--- a/packages/odds-core/odds_core/config.py
+++ b/packages/odds-core/odds_core/config.py
@@ -8,6 +8,8 @@ from pathlib import Path
 from pydantic import Field
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
+from odds_core.sports import SportKey
+
 
 class APIConfig(BaseSettings):
     """The Odds API configuration."""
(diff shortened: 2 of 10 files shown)


Acusick1 (commented):
Reviewed as comment only (Claude-coauthored commits — cannot approve). CI green, tests pass.

## Positive

- `_coerce_bookmaker_list` is self-contained, well-tested, and the comment explicitly flags it as a workaround tied to anthropics/claude-code#22394. Removal trigger is clear.
- `market_hold` move from caller-filtered retail list → all-non-sharp-books-with-per-book-hold is the right call: the agent can filter high-vig books in context without guessing which books are retail.
- `SportKey` on four tool signatures catches the `sport='mlb'` silent-no-op issue the PR description flagged.

## Issues / nits

1. **Per-book hold with missing outcomes (`server.py:966-970`).** `book_hold` sums `calculate_market_hold` over whatever prices a bookmaker has in the snapshot. If a book is missing an outcome (e.g. only home+draw for 1x2), the reported hold is artificially low — possibly negative — with no signal to the caller that it's computed from a partial market. Worth either (a) skipping books with incomplete outcome coverage, or (b) adding an outcome-count field so the agent can filter.

2. **`_DEFAULT_RETAIL_BOOKMAKERS` drift risk (`server.py:56-59`).** Widened from 3 → 5 books with a comment saying it's a placeholder pending the deployed-model TODO. Given `get_event_features` no longer accepts a caller override, this constant is now the only knob. Should probably import from `feature_extraction.py` rather than duplicate, or leave a `# TODO` linking the issue so it doesn't quietly diverge.

3. **`SportKey` partial propagation.** Applied to four MCP tool signatures but `daily_digest.py`, `agent_run.py`, and `config.sports` still use `str`. Not blocking for this PR, but the validation benefit is local until the type propagates into the scheduler job entry points that the MCP tools call.

4. **`_coerce_bookmaker_list` malformed-JSON fallback (`test_malformed_json_falls_back_to_comma_split`).** `"[bet365, betway"` → `["[bet365", "betway"]` is surprising — the opening `[` ends up as part of the first element. Probably fine as workaround behavior, but consider stripping leading/trailing `[]` before comma-split, or raising on malformed JSON so the agent gets a clear error instead of a silently-mangled bookmaker list.

5. **Test coverage gap.** No test asserts `market_hold` appears in `get_sharp_soft_spread` output, or that it's `None` for totals/spreads. Manual verification is in the test plan, but a regression here would not be caught.

Nothing blocking — issues 1 and 4 are the only ones I'd consider addressing in this PR.

Acusick1 (commented):
Re-reviewed the three follow-up commits. All five items addressed — CI green.

## Resolved

1. **Partial-market hold** — `per_book_market_holds` in `odds_analytics/utils.py` drops books missing any required outcome. Helper is cleanly generic (`(bookmaker, outcome, price)` tuples), reusable, and tested in isolation. 
2. **Retail default drift risk** — comment on `_DEFAULT_RETAIL_BOOKMAKERS` now spells out the US-vs-UK divergence from `feature_extraction.DEFAULT_RETAIL_BOOKMAKERS` and flags the drift risk when an EPL model ships.
3. **SportKey propagation** — extended to `JobContext.sport`, `agent_run` (OVERNIGHT_WINDOWS + helpers), `daily_digest` defaults, and `config.Settings.sports`. Pydantic runtime validation on `sports` is the real win — typos in env config will now fail loudly. Adding `basketball_nba` to the literal is pragmatic given the broader NBA support in training/backtest code; the commit message correctly flags that `schedule_next_wakeup` will accept `basketball_nba` but write an orphan wakeup row. Low-consequence, acceptable as-is.
4. **Malformed JSON** — now raises `ValueError` with a clear message; also rejects valid-but-non-list JSON (`"42"`). Test updated.
5. **market_hold test coverage** — three new regression tests cover h2h populated, partial-outcome book gets `None`, totals/spreads get `None`. Plus `test_utils.py::TestPerBookMarketHolds` covers the helper.

## Non-blocking observations

- `scheduling/helpers.py::self_schedule` still takes `sport: str | None` — consistent with the commit message's "SportKey is a str subtype at runtime so callers pass through cleanly" reasoning, so no action needed.
- `SportKey` now has three values but only two are actively scheduled. If a fourth sport lands, a follow-up splitting `SportKey` (canonical set) from `ScheduledAgentSportKey` (sports the scheduler actually handles) is worth considering — already flagged in the commit message.

Nothing else. Still comment-only since commits remain Claude co-authored — approval is yours to grant.

PR #333, opened 2026-04-18
Parameterize OddsPortal over/under line for MLB totals

Acusick1:
## Summary

MLB totals support in two layers:

**1. Converter parameterization.** The upcoming-scrape adapter hardcoded `point: 2.5`, so any non-football line (MLB's 7.5/8/8.5/9+, quarter lines) was silently rejected as "Unsupported market". Shared `parse_over_under_line()` in `oddsportal_common`; `convert_upcoming_matches` dispatches any `over_under_X_Y` dynamically and threads the parsed line into `_convert_over_under_match(line=...)`. Default stays 2.5 for the EPL path. `scripts/ingest_oddsportal.py` now reuses the helper instead of its duplicate regex.

**2. Main-line discovery via ESPN (no paid API).** Scraping all 11 enum values per match would ~10× per-game Playwright cost. Instead, `espn_mlb_odds.get_mlb_main_totals()` fetches the public ESPN scoreboard + per-event odds endpoint, reads the root `overUnder` (DraftKings' featured line, populated reliably from T-12h onward), and returns the distinct lines seen across the slate — typically 2–4. `odds scrape upcoming --auto-totals` augments the MLB `LeagueSpec.markets` with those lines before calling `ingest_league`. ESPN failures degrade gracefully to the base markets.

ESPN is undocumented and single-book; treated as fragile infrastructure (per-event try/except, empty fallback, no hard dependency). Scheduled MLB totals scraping (wiring `--auto-totals` into the periodic job) is out of scope here — this gives the agent/CLI on-demand access first, before we pay the scheduled scrape cost.

## Test plan

- [x] `uv run pytest tests/unit/test_oddsportal_adapter.py` (66 pass)
- [x] `uv run pytest tests/unit/test_espn_mlb_odds.py` (16 pass; mocked happy path + network failure + malformed shape + session lifecycle)
- [x] `uv run pytest tests/unit` (1837 pass; only pre-existing `test_settings_defaults` Discord-leak failure)
- [x] `uv run ruff check` / `ruff format` clean
- [ ] Manual: `uv run odds scrape upcoming --league mlb --auto-totals --dry-run` against a live MLB slate

### DIFF
--- a/packages/odds-cli/odds_cli/commands/scrape.py
+++ b/packages/odds-cli/odds_cli/commands/scrape.py
@@ -4,11 +4,14 @@ from __future__ import annotations
 
 import asyncio
 import json
-from typing import Annotated, Any
+from typing import TYPE_CHECKING, Annotated, Any
 
 import typer
 from rich.console import Console
 from rich.table import Table
+
+if TYPE_CHECKING:
+    from odds_lambda.jobs.fetch_oddsportal import LeagueSpec
 
 app = typer.Typer()
 console = Console()
@@ -26,9 +29,17 @@ def scrape_upcoming(
     from_file: str | None = typer.Option(
         None, "--from-file", help="Load matches from JSON file instead of scraping"
     ),
+    auto_totals: bool = typer.Option(
+        False,
+        "--auto-totals",
+        help=(
+            "MLB only: query ESPN for each game's featured Over/Under line and "
+            "add the distinct over_under_X_Y markets to the scrape."
+        ),
+    ),
 ) -> None:
     """Scrape upcoming match odds from OddsPortal and ingest into the pipeline."""
-    asyncio.run(_scrape_upcoming(league, market, dry_run, from_file))
+    asyncio.run(_scrape_upcoming(league, market, dry_run, from_file, auto_totals))
 
 
 async def _scrape_upcoming(
@@ -36,6 +47,7 @@ async def _scrape_upcoming(
     markets: list[str] | None,
     dry_run: bool,
     from_file: str | None,
+    auto_totals: bool,
 ) -> None:
     import dataclasses
 
@@ -54,6 +66,9 @@ async def _scrape_upcoming(
     spec = LEAGUE_SPEC_BY_NAME[league]
     if markets:
         spec = dataclasses.replace(spec, markets=markets)
+
+    if auto_totals:
+        spec = await _augment_spec_with_espn_totals(spec)
 
     raw_matches: list[dict[str, Any]] | None = None
     if from_file:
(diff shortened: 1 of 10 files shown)


Acusick1 (commented):
Reviewed. Clean refactor — removing the static `over_under_2_5` mapping in favour of dispatching on `parse_over_under_line()` is the right call, and pulling the helper into `oddsportal_common` (plus deleting the duplicate regex in `scripts/ingest_oddsportal.py`) removes real duplication. ESPN module is appropriately paranoid: per-event try/except, graceful degradation when the scoreboard fails, session-lifecycle handled, and no hard dependency introduced for the MLB path. CI green (24603852851), 1837 unit tests pass.

A few things worth flagging — none blocking, most are risk-tolerance calls:

**Lexicographic sort on `over_under_X_Y` keys.** `distinct_market_keys` sorts strings, so a realistic MLB slate with 9.5, 10.5, 11.5 lines yields `["over_under_10_5", "over_under_11_5", "over_under_9_5"]`. OddsHarvester likely doesn't care about order, but the log line `"distinct lines: ..."` will read oddly. Trivial fix if you want it: sort by the parsed float.

**`overUnder = 0.0` isn't filtered.** The docstring correctly calls out that `current.total.value` is unreliable and often 0.0, but the root `items[0].overUnder` falls through the `isinstance(line, int | float)` check with value 0.0 if ESPN ever returns it for a pre-release game. `line_to_market_key(0.0)` would produce `over_under_0_0`, which OddsHarvester would presumably reject — so harmless, but a `line > 0` guard would be cheap insurance.

**`isinstance(line, int | float)` accepts `bool`.** bool subclasses int in Python. Not a realistic ESPN response shape, but strict typing would exclude it.

**Manual smoke test unchecked.** PR body's final checklist item (`--auto-totals --dry-run` against live MLB slate) is still unchecked. Given ESPN is undocumented and the assumption "overUnder populated reliably from T-12h onward" is load-bearing, running it against tomorrow's slate before merging feels warranted — you'll also validate the sorted-log cosmetic question above without writing more code.

**Test gaps (minor).** No test exercises the `target_date` parameter's `YYYYMMDD` serialization to the `dates` query param, and no test covers `convert_upcoming_matches` still raising `ValueError` for an unknown non-OU market (the dispatch ordering changed — parse first, fall back to map — so it's worth one regression guard).

Out-of-scope note / sanity check for the roadmap: wiring `--auto-totals` into the scheduled fetch_oddsportal job (deferred per the PR body) will need to decide how often to re-query ESPN since lines move intraday — probably recompute per scrape tick rather than caching the opening discovery. Worth capturing as a follow-up issue.

Cannot approve — PR commits are co-authored by Claude.

PR #335, opened 2026-04-18
Tighten OddsPortal event match window to ±2h to prevent conflation

### LINKED ISSUE #334: Fix find_or_create_event window and re-point conflated MLB snapshots
## Goal

`find_or_create_event` uses a ±24h team+date match window, so back-to-back same-matchup games (MLB series pattern) get silently merged: the second day's scrape writes odds onto the first day's event row. Local DB shows 38 MLB events with 418 misattributed snapshots (~56% of MLB snapshots). EPL/NBA are incidentally unaffected. Production has the same bug and needs both a writer fix and a data migration.

## Scope

- Tighten the team+date matching window in `find_or_create_event` from ±24h to ±2h.
- Add an Alembic data migration that detects conflated snapshots and re-points them to correct event rows (creating new event rows where needed).
- Update/extend writer tests to cover: true idempotent re-scrape, same-teams on consecutive days (previously broken), same-day doubleheader game 2.

## Success criteria

- [ ] `find_or_create_event` no longer matches an event more than ±2h from the supplied `match_date`.
- [ ] Migration re-points every snapshot where `|event.commence_time − (snapshot_time + hours_until_commence)| > 2h AND hours_until_commence > 0` to an event row whose `commence_time` matches within ±2h, creating the row if absent.
- [ ] Post-migration, running the divergence query (below) returns 0 rows on local dev DB.
- [ ] Migration runs cleanly on production DB without downtime.
- [ ] New writer test covers: team A vs B on day N and day N+1 as distinct events (series-day case).
- [ ] New writer test covers: team A vs B at 13:00 and 18:00 same day as distinct events (doubleheader case).

## Location

- Writer: `packages/odds-lambda/odds_lambda/storage/writers.py:84-145`
- Migration: new file under `migrations/versions/`
- Tests: `tests/unit/test_team_normalization.py` (or new `tests/unit/test_find_or_create_event.py` if cleaner)

## Implementation approach

- Change `timedelta(hours=24)` → `timedelta(hours=2)` at `writers.py:100-101`. Keep the rest of the matching logic (team normalization, `>1 candidate` warning path) unchanged.
- Migration algorithm:
  1. Select snapshots where `hours_until_commence > 0` and `|commence_time − (snapshot_time + make_interval(hours => hours_until_commence))| > interval '2 hours'`.
  2. For each, compute `true_commence = snapshot_time + hours_until_commence`.
  3. Find an event row with same `(home_team, away_team, sport_key)` and `commence_time` within ±2h of `true_commence`. If none, insert a new event row using `_build_event_id(home, away, true_commence)` and `status = SCHEDULED`.
  4. Update `odds_snapshots.event_id` to the correct row.
  5. Do not delete the original event rows — they still carry their own legitimate snapshots (from the earlier day they were scraped before the next day's run overwrote them).
- Migration should be idempotent: re-running after the writer fix is deployed should be a no-op.

## Constraints

- `fetch_tier` and `hours_until_commence` on affected snapshots are already correct relative to the true match (they were computed from the scraped `match.match_date`, not the event row) — do not recompute them.
- Use async Alembic patterns consistent with existing migrations in `migrations/versions/`.

## Notes

- Verification query for success criteria: `SELECT COUNT(*) FROM odds_snapshots os JOIN events e ON e.id = os.event_id WHERE os.hours_until_commence > 0 AND ABS(EXTRACT(EPOCH FROM (e.commence_time - (os.snapshot_time + (os.hours_until_commence || ' hours')::interval))) / 3600) > 2;`
- Odds API ingestion path is not affected (uses `bulk_upsert_events` with API-provided IDs). Only the OddsPortal scraper path (`fetch_oddsportal.py:178`) hits this bug.
- MLB training datasets don't yet consume `odds_snapshots`, so no downstream model retraining is required — this is purely a data-integrity fix.
- Doubleheader game 2 typically starts 3+ hours after game 1; ±2h comfortably separates them.

Acusick1:
## Summary
- Narrowed `find_or_create_event` match window from ±24h to ±2h so MLB series (consecutive-day same-matchup games) no longer get conflated onto a single `Event` row.
- Added `THHMM` suffix to the deterministic `op_live_*` event id so same-day doubleheaders mint distinct rows; existing rows are untouched and the ±2h SQL window still locates them on re-scrape.
- Shipped an Alembic data migration (`cc937c8cd362`) that repoints divergent snapshots, child `odds` rows, `predictions`, `paper_trades`, and `match_briefs` to the correct event (creating new events where needed). Sport-agnostic but logs a per-sport blast-radius breakdown before any mutation, plus a sanity check for any overlapping odds timestamps. Idempotent on re-run.
- Added unit tests covering series-day, doubleheader, and re-scrape idempotency cases.

## Closes #334

### DIFF
--- a/migrations/versions/cc937c8cd362_repoint_conflated_oddsportal_snapshots.py
+++ b/migrations/versions/cc937c8cd362_repoint_conflated_oddsportal_snapshots.py
@@ -55,6 +55,11 @@ def _build_event_id(home_team: str, away_team: str, commence_time: datetime) -> 
 
 # Snapshots where the event row's commence_time is more than 2h away from
 # the true match time implied by (snapshot_time + hours_until_commence).
+#
+# Intentionally sport-agnostic: the >2h divergence signature is valid across
+# sports, and the correctness of the repoint does not depend on sport_key.
+# The upgrade() function logs a per-sport breakdown so the blast radius is
+# explicit when the migration runs.
 _DIVERGENCE_QUERY = sa.text(
     """
     SELECT
@@ -62,7 +67,7 @@ _DIVERGENCE_QUERY = sa.text(
         os.event_id AS old_event_id,
         os.snapshot_time,
         os.hours_until_commence,
-        os.snapshot_time + make_interval(secs => os.hours_until_commence * 3600)
+        os.snapshot_time + make_interval(hours => os.hours_until_commence)
             AS true_commence,
         e.home_team,
         e.away_team,
@@ -76,11 +81,63 @@ _DIVERGENCE_QUERY = sa.text(
           EXTRACT(
               EPOCH FROM (
                   e.commence_time
-                  - (os.snapshot_time + make_interval(secs => os.hours_until_commence * 3600))
+                  - (os.snapshot_time + make_interval(hours => os.hours_until_commence))
               )
           ) / 3600
       ) > 2
     ORDER BY os.snapshot_time
+    """
+)
+
+
+# Per-sport breakdown of the divergence set, logged before repointing runs so
+# the blast radius is explicit in migration output.
+_DIVERGENCE_BREAKDOWN_QUERY = sa.text(
+    """
+    SELECT e.sport_key, COUNT(*) AS n
+    FROM odds_snapshots os
+    JOIN events e ON e.id = os.event_id
+    WHERE os.hours_until_commence IS NOT NULL
+      AND os.hours_until_commence > 0
+      AND ABS(
+          EXTRACT(
+              EPOCH FROM (
+                  e.commence_time
+                  - (os.snapshot_time + make_interval(hours => os.hours_until_commence))
+              )
+          ) / 3600
+      ) > 2
+    GROUP BY e.sport_key
+    ORDER BY n DESC
+    """
+)
+
+
+# Sanity check for the uniqueness assumption embedded in _REPOINT_ODDS: that
+# (event_id, odds_timestamp) uniquely identifies the odds rows belonging to
+# one snapshot, so UPDATE needs no snapshot_id tiebreaker. This holds because
+# each scrape run writes exactly one snapshot per event with a single
+# snapshot_time value shared by all its odds rows. The query below flags any
+# event where two or more snapshots share the same snapshot_time within the
+# divergent set — that would indicate overlapping scrapes and mean the UPDATE
+# could drag unrelated odds rows along with the one we intend to move.
+_ODDS_TIMESTAMP_SANITY_QUERY = sa.text(
+    """
+    SELECT os.event_id, os.snapshot_time, COUNT(*) AS n
+    FROM odds_snapshots os
+    JOIN events e ON e.id = os.event_id
+    WHERE os.hours_until_commence IS NOT NULL
+      AND os.hours_until_commence > 0
+      AND ABS(
+          EXTRACT(
+              EPOCH FROM (
+                  e.commence_time
+                  - (os.snapshot_time + make_interval(hours => os.hours_until_commence))
+              )
+          ) / 3600
+      ) > 2
+    GROUP BY os.event_id, os.snapshot_time
+    HAVING COUNT(*) > 1
     """
 )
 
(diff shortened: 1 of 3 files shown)


Acusick1 (commented):
## Review — comment only (PR is co-authored by Claude)

CI green, writer change is clean, unit tests cover the three cases that matter (series-day, doubleheader, re-scrape idempotency). Overall the approach is sound. One blocker and a couple of smaller concerns with the data migration below.

### Blocker: `make_interval(hours => double precision)` will error on production

`odds_snapshots.hours_until_commence` is stored as `double precision` (Python `float`, with fractional values like `7.5`, `0.5`, `47.33`). Postgres does not overload `make_interval` for double precision on the `hours` kwarg — only `secs` accepts double precision. Verified on Postgres 14:

```
postgres=# CREATE TEMP TABLE t (h double precision);
postgres=# INSERT INTO t VALUES (8.5), (3.0);
postgres=# SELECT make_interval(hours => h) FROM t;
ERROR:  function make_interval(hours => double precision) does not exist
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

So the three queries that use `make_interval(hours => os.hours_until_commence)` — `_DIVERGENCE_QUERY`, `_DIVERGENCE_BREAKDOWN_QUERY`, `_ODDS_TIMESTAMP_SANITY_QUERY` — will raise as soon as they touch any row with a fractional `hours_until_commence`. Which, given the real data, is essentially every row.

The issue's verification query already uses the correct form: `(os.hours_until_commence || ' hours')::interval`. Suggest switching all three queries to that (or `os.hours_until_commence * interval '1 hour'`). As far as I can tell, this migration has not actually been applied anywhere yet — the CI job runs `test_schema_validation` but doesn't run `alembic upgrade` with real data, and the 38-conflated-event figure was from the raw SELECT query, not from a dry run of the migration. Worth running the migration end-to-end against a local copy before deploying.

### Non-blocking: sanity check is non-fatal

`_ODDS_TIMESTAMP_SANITY_QUERY` warns and continues. If the uniqueness assumption behind `_REPOINT_ODDS` is violated, the migration will silently move unrelated odds rows. Given the rest of the migration is explicit about blast radius, I'd lean toward `raise RuntimeError(...)` if any duplicates come back, and require human review before applying.

### Non-blocking: paper_trades / match_briefs repoint is coarse-grained

`_REPOINT_PAPER_TRADES` and `_REPOINT_MATCH_BRIEFS` move every row from `old_event_id` to `new_event_id`. But the \"old\" event still owns its legitimate day-1 snapshots — so a trade/brief placed on those day-1 snapshots would incorrectly follow the day-2 snapshots to the new event. `paper_trades` has no `snapshot_id` column, so we can't distinguish. In practice this is probably a no-op (EPL is unaffected per the issue, MLB agent isn't live), but I'd verify with a quick pre-flight `SELECT COUNT(*) FROM paper_trades WHERE event_id IN (<affected>)` and short-circuit if it's zero, rather than silently moving them. Or at minimum, log which rows are being moved so the blast radius is auditable.

### Minor: sport-consistency between writer and migration

`_FIND_EVENT_QUERY` filters by `sport_key`; `find_or_create_event` in the writer does not. Not a correctness issue given team names are sport-disjoint in practice, but worth aligning (probably by adding `sport_key` to the writer's match filter) so the two paths can't diverge on teams-only collisions in the future.

### Minor: `_EVENT_EXISTS_QUERY` reuse without verification

In the `existing is None` branch, if `_build_event_id` happens to collide with a row that isn't within the ±2h window, the migration reuses that id without re-verifying teams/sport/commence. Rare, but an `assert` would be cheap insurance.

Happy to approve once the `make_interval` issue is resolved and the migration has been exercised end-to-end against real data.

Acusick1:
Review addressed in `5f1b72a`:

**Blocker fix**
- Replaced `make_interval(hours => os.hours_until_commence)` with `(os.hours_until_commence || ' hours')::interval` in `_DIVERGENCE_QUERY`, `_DIVERGENCE_BREAKDOWN_QUERY`, `_ODDS_TIMESTAMP_SANITY_QUERY`. Confirmed locally that the original form raises the expected `function make_interval(hours => double precision) does not exist` against real data — the bug was latent because CI validates the chain against an empty schema.

**End-to-end migration run (local Postgres 15)**
- `alembic stamp 917adbe0576d` → `alembic upgrade head`: clean. Logs:
  - `Repointing 0 snapshots: no divergent rows found`
  - `paper_trades: 0 rows to repoint`
  - `match_briefs: 0 rows to repoint`
- Issue's verification query returns 0 divergent rows within the migration's scope (`hours_until_commence > 0`). 196 rows with `hours_until_commence = 0.0` are intentionally out of scope — no true commence time is reconstructable for those.
- Re-run upgrade is a no-op.

**Other review points**
- Sanity check now `raise RuntimeError(...)` with up to 5 offending `(event_id, snapshot_time)` pairs instead of warn-and-continue.
- `paper_trades` / `match_briefs` repoint gets a pre-flight `COUNT(*)` over the full `old_id` set. If 0, logs and skips all UPDATEs. If >0, logs the count plus up to 5 `(id, event_id)` sample rows before running the UPDATEs.
- `find_or_create_event` now filters by `sport_key` (one-line plumbing — already a function parameter) so the writer and migration paths can't diverge on teams-only collisions.
- `_EVENT_EXISTS_QUERY` now selects `home_team`, `away_team`, `sport_key`, `commence_time`. In the `existing is None` branch, if the deterministic id collides with an outside-window row, the migration raises `RuntimeError` on mismatch instead of reusing.

**Note for operators**
If any environment was previously stamped at `cc937c8cd362` against data, the old `make_interval` query would have errored on first data row — so it almost certainly never ran to completion. A follow-up `alembic stamp cc937c8cd362^` + `alembic upgrade head` would be needed to actually execute the repoint there.

CI green on the latest push.

Acusick1 (commented):
Thanks — all four points resolved:

- **make_interval → `(col || ' hours')::interval`** in all three queries. ✅
- **Sanity check is now a hard `RuntimeError`** with a 5-pair sample. ✅
- **paper_trades / match_briefs** get a pre-flight count + sample logged, and the UPDATEs are short-circuited when the count is zero. The fundamental ambiguity (no snapshot_id on `paper_trades`, so we can't tell which trades belong to which day's snapshots) is unresolved, but in practice the count should be zero on production (EPL unaffected, MLB agent not live), and the pre-flight sample will surface it if that assumption is ever wrong. Acceptable. ✅
- **`_EVENT_EXISTS_QUERY` collision path** now verifies teams/sport/commence_time match before reusing, raises with a diff'd message if not. ✅
- **Writer now filters on `sport_key`** so it can't diverge from the migration's `_FIND_EVENT_QUERY`. ✅

End-to-end dry-run note in the commit message (stamp down → upgrade → verification query returns 0 → re-upgrade is no-op) is exactly what I was looking for.

CI green. LGTM — happy for this to go in once you're ready. (Leaving as a comment rather than approve since I'm co-author.)

PR #338, opened 2026-04-20
Reject stale OddsPortal upcoming scrapes and stop masking tier fields

Acusick1:
## Summary

The MLB OddsPortal scrape has been silently writing live upcoming-game odds against ghost event rows with historical commence_times (e.g. `op_live_BALORI_KANROY_2024-10-02T2038`, `op_live_TORJAY_LOSANG_2025-07-06T1737`).

**Root cause:** OddsPortal's baseball league page links to `/baseball/h2h/...` pages. On those pages, `#react-event-header` intermittently hydrates with the most recent **past** meeting between the two teams. OddsHarvester reads that stale header (wrong `match_date`, swapped home/away) while the market widgets on the same page return today's live odds — producing frankenstein records.

**Verified:** 888sport stored "2024-10-02" BAL/KC last snapshot was -110 / -118, identical to today's live KC @ BAL line pulled via Playwright. Actual 2024-10-02 ALWC Game 2 closing was BAL -150 / KC +128, so stored odds cannot be historical. Same pattern on 2025-07-06 TOR vs LAA: real closing was TOR -189 / LAA +157, stored was a near pick-em line matching today's upcoming fixture.

## Fixes

**1. Adapter-side stale-match filter** in `convert_upcoming_matches`. Rejects any match whose `match_date < scraped_date - 2h`, logs a `stale_match_rejected` warning with home/away/hours_in_past/match_link. An upcoming scrape should never surface a genuinely finished game; this is boundary validation against the third-party harvester.

**2. Remove the silent overwrite** in `ingest_league` that stomped `hours_until_commence` and `fetch_tier` with values recomputed from `match.match_date`. `store_odds_snapshot` already computes both correctly from the event's `commence_time`. The overwrite was:
- Collapsing legitimate in-play snapshots (negative hours) to `closing / 0`
- Hiding frankenstein writes from the #335 repoint migration's `hours_until_commence > 0` predicate

## Tradeoff

On a scrape run that hits the hydration bug, the filter drops the affected matches and we lose OddsPortal coverage for those fixtures until the next scrape cycle (30–60 min). This is preferable to silent corruption: ghost events with wrong-team home/away leaked into feature extraction, backtests, and paper-trade lookups. The `stale_match_rejected` log fires to Discord so frequency is observable.

## Follow-ups (separate PRs)

- Cleanup migration for existing ghost events on dev and prod
- Root cause fix in OddsHarvester (wait for `#react-event-header` to hydrate to upcoming fixture before reading)
- Optional guardrail in `find_or_create_event` to refuse minting new events with `commence_time < now - 1h`

## Test plan

- [x] New adapter tests: stale match rejected, match within threshold accepted, mixed good+stale batch only drops stale
- [x] 57 adapter tests + 25 fetch-oddsportal tests pass
- [x] 322 related unit tests (oddsportal / ingest / fetch) pass
- [x] Ruff lint + format clean
- [ ] Observe `stale_match_rejected` Discord alerts post-deploy to gauge bug frequency
- [ ] Confirm no new ghost events (`op_live_*` with pre-today `commence_time`) created after deploy

### DIFF
--- a/packages/odds-lambda/odds_lambda/jobs/fetch_oddsportal.py
+++ b/packages/odds-lambda/odds_lambda/jobs/fetch_oddsportal.py
@@ -30,7 +30,7 @@ from odds_lambda.oddsportal_adapter import (
     MatchOdds,
     convert_upcoming_matches,
 )
-from odds_lambda.oddsportal_common import hours_to_tier, run_scraper_with_retry
+from odds_lambda.oddsportal_common import run_scraper_with_retry
 from odds_lambda.scheduling.helpers import apply_overnight_skip, get_next_kickoff, self_schedule
 from odds_lambda.scheduling.jobs import JobContext, make_compound_job_name
 from odds_lambda.storage.writers import OddsWriter
@@ -195,9 +195,6 @@ async def ingest_league(
                 )
 
                 snapshot.api_request_id = api_request_id
-                hours_before = (match.match_date - match.scraped_date).total_seconds() / 3600
-                snapshot.hours_until_commence = max(0.0, hours_before)
-                snapshot.fetch_tier = hours_to_tier(hours_before)
                 stats.snapshots_stored += 1
 
             except Exception as e:
--- a/packages/odds-lambda/odds_lambda/oddsportal_adapter.py
+++ b/packages/odds-lambda/odds_lambda/oddsportal_adapter.py
@@ -9,8 +9,10 @@ from __future__ import annotations
 import re
 from collections.abc import Callable
 from dataclasses import dataclass
-from datetime import UTC, datetime
+from datetime import UTC, datetime, timedelta
 from typing import Any
+
+import structlog
 
 from odds_lambda.oddsportal_common import (
     DRAW_OUTCOME,
@@ -18,6 +20,16 @@ from odds_lambda.oddsportal_common import (
     normalize_bookmaker_key,
     parse_match_date,
 )
+
+logger = structlog.get_logger()
+
+# Reject upcoming-scrape matches whose claimed start time is already this far
+# in the past at scrape time. Guards against a known OddsPortal bug where the
+# H2H page's React event header occasionally hydrates with the most recent
+# PAST meeting between the two teams (stale `startDate` + swapped home/away),
+# producing frankenstein records — live odds pinned to a historical date. An
+# upcoming-scrape should never surface a genuinely finished game.
+STALE_MATCH_REJECT_THRESHOLD = timedelta(hours=2)
 
 # Regex: Betfair format is "FRAC_REPEAT(LIQUIDITY)" e.g. "99/10099/100(300)"
 # The fraction is repeated (concatenated), followed by optional (liquidity).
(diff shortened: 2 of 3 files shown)


Acusick1 (commented):
Reviewed. Verified both fixes land cleanly:

**Fix 1 (stale filter)** — `convert_upcoming_matches` correctly rejects `match_dt < scraped_dt - 2h` with a structured `stale_match_rejected` log including `match_link` for root-cause investigation. Test coverage hits rejection, boundary acceptance, and mixed-batch isolation.

**Fix 2 (overwrite removal)** — confirmed against `OddsWriter.store_odds_snapshot` (writers.py:186-190) and `tier_utils.calculate_tier_from_timestamps`: the writer already derives `fetch_tier` and `hours_until_commence` from `event.commence_time` + `snapshot_time`, which is authoritative. The removed block was recomputing from `match.match_date - match.scraped_date`, which diverges from the true event clock whenever `find_or_create_event` resolves to an existing event (e.g. in the frankenstein case). Removing it closes the cover-up for #335's predicate.

**Questions / nits:**

1. **2h threshold rationale** — not explained in the PR. For baseball, first-pitch + 2h occasionally overlaps with legitimately late-running games (extra innings push past 3-4h). If OddsPortal ever surfaces such matches on its upcoming endpoint, we'd silently drop them. Given the observed bug is *days-to-months* in the past, a wider threshold (e.g. 12h) would catch the same frankenstein cases while leaving more margin for genuine late in-play edge cases. Not blocking — the tradeoff note covers it — but worth leaving a comment or constant-level docstring justifying the 2h choice so it doesn't get tightened further without thought.

2. **Follow-up coverage** — the PR lists a `find_or_create_event` guardrail (refuse minting events with `commence_time < now - 1h`) as optional. Given the root cause is third-party and the fix here is boundary-level, I'd bump that to non-optional: it's a second independent layer against the same class of bug (any caller, not just OddsPortal, is prevented from minting retroactive events). Worth filing as its own issue even if the guardrail PR is deferred.

3. **Observability** — `stale_match_rejected` fires at `warning` level. Is that enough to surface in Discord given the PR's post-deploy checklist item depends on observing frequency? Confirm the alert filter captures warning-level entries from this logger, otherwise it'll be invisible.

4. **Cleanup migration** — mentioned as a separate PR. Worth filing the issue now so the ghost rows don't linger indefinitely in dev/prod — they'll continue polluting feature extraction / briefs until removed.

Scope, tests, type hints, timezone handling (UTC throughout), and structlog usage all match codebase conventions. LGTM once the rationale + follow-up issues are tracked.

Acusick1:
Addressed review:

**(1) Threshold widened 2h → 12h** (d8f1d7b). Expanded constant docstring with rationale — observed frankenstein dates are days-to-months old, so any >3-4h catches them; 12h leaves margin for legitimately long in-play scrapes so we never drop a game still in progress.

**(2) Guardrail filed as #340.** Scoped the backfill-script concern: proposed applying the check only to events with the \`op_live_\` id prefix (the live-scrape caller) so historical ingestion paths are unaffected.

**(3) Observability — you were right, my PR description was wrong.** Verified \`job_alert_context\` only fires on exceptions / heartbeats; \`logger.warning()\` calls are not forwarded to Discord — they go to CloudWatch via structlog. Post-deploy observability is via CloudWatch log queries for \`stale_match_rejected\`, not Discord. Happy to add aggregated Discord alerting in this PR (count rejects in \`ingest_league\` → \`send_job_warning\` if > 0, rate-limited) if you'd prefer — otherwise defer.

**(4) Cleanup migration filed as #339.** Listed known affected event ids and the predicate (\`commence_time < created_at - 6h\` on \`op_live_%\` events). Shape mirrors cc937c8cd362.

PR #344, opened 2026-04-20
Refuse to mint stale op_live_* events at the writer layer

### LINKED ISSUE #340: Guardrail: refuse to mint live-scrape events with past commence_times
## Context

Follow-up to PR #338. The adapter-level stale-match filter catches the known OddsPortal frankenstein pattern at one boundary. A second independent layer at \`find_or_create_event\` would catch any caller — not just OddsPortal — silently minting retroactive \`op_live_*\` events.

## Proposed rule

In \`OddsWriter.find_or_create_event\`, before creating a new \`Event\` row, reject if \`match_date < datetime.now(UTC) - 1h\` (or similar window). Raise \`ValueError\` with home/away/match_date so the call site sees it.

## Scoping concern

Manual backfill scripts (e.g. historical OddsPortal results ingestion, football-data.co.uk import) legitimately create past-commence events. A blanket reject would break them. Two options:

**A.** Scope the guardrail to event id prefix — only events minted with the \`op_live_\` prefix (i.e. live upcoming scrapes) get the check. Historical ingestion uses different id schemes.

**B.** Add a \`source\` / \`creation_mode\` parameter that backfill callers pass explicitly (\`creation_mode=\"historical\"\`) to bypass the guard.

I lean toward **A**: the id prefix already encodes the caller's intent, and no other code path should be minting \`op_live_*\` with past dates.

## Testing

- Accept: \`match_date\` within ±1h of now.
- Reject: \`match_date\` > 1h in the past (for live-prefixed ids).
- Allow: historical ingestion paths (different id prefix).

## Priority

Defense in depth — not blocking. Low effort if scoped via option A.

Acusick1:
## Summary
- Adds a defense-in-depth guardrail in `OddsWriter.find_or_create_event`: rejects creating a new `Event` row when the event id starts with `op_live_` and `match_date` is more than 1 hour in the past.
- Scoped via id prefix (Option A from the issue) — historical backfill paths use different id schemes and remain unaffected.
- Raises `ValueError(home, away, match_date)` so the per-match `try/except` in `fetch_oddsportal.py` records it in `stats.errors` without aborting the run.
- Constants `LIVE_EVENT_ID_PREFIX` and `LIVE_EVENT_STALE_THRESHOLD` defined once on `OddsWriter` and reused.

## Tests
- Accept: live-prefix event within ±1h of now.
- Reject: live-prefix event with `match_date` 2h in the past raises `ValueError`.
- Idempotent: pre-existing stale `op_live_*` row is found, not re-created.
- Bypass: historical-prefix events created via `upsert_event` are unaffected.

## Closes #340

### DIFF
--- a/packages/odds-lambda/odds_lambda/storage/writers.py
+++ b/packages/odds-lambda/odds_lambda/storage/writers.py
@@ -21,6 +21,9 @@ logger = structlog.get_logger()
 class OddsWriter:
     """Handles all write operations to the database."""
 
+    LIVE_EVENT_ID_PREFIX = "op_live_"
+    LIVE_EVENT_STALE_THRESHOLD = timedelta(hours=1)
+
     def __init__(self, session: AsyncSession):
         """
         Initialize writer with database session.
@@ -72,9 +75,6 @@ class OddsWriter:
             self.session.add(event)
             logger.info("event_created", event_id=event.id)
             return event
-
-    LIVE_EVENT_ID_PREFIX = "op_live_"
-    LIVE_EVENT_STALE_THRESHOLD = timedelta(hours=1)
 
     @staticmethod
     def _build_event_id(home_team: str, away_team: str, match_date: datetime) -> str:
--- a/packages/odds-lambda/odds_lambda/storage/writers.py
+++ b/packages/odds-lambda/odds_lambda/storage/writers.py
@@ -73,6 +73,9 @@ class OddsWriter:
             logger.info("event_created", event_id=event.id)
             return event
 
+    LIVE_EVENT_ID_PREFIX = "op_live_"
+    LIVE_EVENT_STALE_THRESHOLD = timedelta(hours=1)
+
     @staticmethod
     def _build_event_id(home_team: str, away_team: str, match_date: datetime) -> str:
         """Generate deterministic event ID for OddsPortal-sourced events.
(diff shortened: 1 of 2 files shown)


Acusick1 (commented):
Reviewed locally on `issue-340-guardrail-stale-live-events`. Ran the 4 new guardrail tests — all pass. CI is green.

**Looks good:**
- Matches Option A from #340: scoping via `op_live_` prefix keeps historical backfill (FDUK, etc., via `upsert_event`) unaffected, confirmed by `test_historical_ingestion_via_upsert_event_unaffected`.
- Order of checks in `find_or_create_event` is correct: the ±2h team+date window (lines 111–121) and the exact-id existence check (140–142) both run before the guardrail, so re-scrapes of known stale events are looked up rather than rejected — `test_finds_existing_stale_live_event_without_creating` pins this.
- `ValueError` is surfaced to the per-match `try/except Exception` at `packages/odds-lambda/odds_lambda/jobs/fetch_oddsportal.py:200-203`, so a stale mint lands in `stats.errors` without aborting the whole ingest — behaves as described.
- `_build_event_id` now references `OddsWriter.LIVE_EVENT_ID_PREFIX` so prefix and guardrail can't drift.
- `ensure_utc(match_date)` on the stale-cutoff comparison is the right call given naive datetimes occasionally show up from adapter code.

**Minor observations (non-blocking):**
- 1h threshold is hardcoded on the class. Fine for now; only worth revisiting if a legit use case appears.
- Defense-in-depth by design, so this won't typically fire (PR #338's adapter filter is the primary boundary). Worth watching Discord `match_ingestion_failed` logs post-deploy to confirm it stays quiet, and to catch any unexpected caller that starts hitting it.

Co-authored by Claude, so leaving as comment rather than approve.

PR #345, opened 2026-04-20
Fix fetch-oddsportal-results DB session scope and self-schedule reliability

### LINKED ISSUE #343: Fix fetch-oddsportal-results Neon idle-timeout during long OddsHarvester scrape
## Goal

The `fetch-oddsportal-results` scraper Lambda is currently non-functional for any sport with past-commence pending events. It acquires a DB session before a 6–8 minute OddsHarvester scrape, Neon closes the idle connection during the scrape, and the post-scrape DB writes crash with `InterfaceError: connection is closed` → `PendingRollbackError`. The job's exception handler re-raises before reaching its self-schedule call, so the EventBridge rule stays `DISABLED` and the job never runs again. Observed on `odds-fetch-oddsportal-results-epl` on prod as of 2026-04-20 (rule DISABLED since deploy, 5 consecutive failed invocations in CloudWatch).

## Scope

- Restructure `process_results()` in `packages/odds-lambda/odds_lambda/jobs/fetch_oddsportal_results.py` so no DB session is held open across the OddsHarvester scrape.
- Fetch pending events into memory, close session. Run scrape with no open session. Open a fresh session for the write phase.
- Ensure the write phase is resilient to per-record failures (one bad record shouldn't kill the batch or prevent self-scheduling).
- Verify `main()` reliably reaches its `schedule_next_execution` call after `process_results()` returns OR raises.

## Location

- `packages/odds-lambda/odds_lambda/jobs/fetch_oddsportal_results.py` — `process_results()` (lines 146–271) and `main()` (lines 274–311)

## Implementation approach

Split `process_results()` into three phases with distinct session scopes:

1. **Read phase** — open session, read pending events into a plain `list[Event]` (detach or copy the data needed — id, home_team, away_team, commence_time), close session.
2. **Scrape phase** — run `run_harvester_historic(spec)` with no open session. This is the long-duration part (6–8+ min) and holds no DB resources.
3. **Write phase** — open a fresh session, iterate records, `update_event_status` + `store_odds_snapshot` per match, commit at the end (or per-record if batch-level commit is risky). Wrap per-record work in its own try/except so one bad record doesn't kill the batch.

`main()` should always attempt self-schedule regardless of whether `process_results()` raised — move the scheduling call into a `try/finally` or pull it out of `job_alert_context` so a results-processing crash doesn't leave the rule disabled.

## Success criteria

- [ ] `process_results()` holds no open DB session during the OddsHarvester scrape
- [ ] `fetch-oddsportal-results-epl` rule returns to `State=ENABLED` with a future schedule after the job runs once on prod
- [ ] A forced error in one record's update path does not prevent the job from self-scheduling
- [ ] CloudWatch shows `fetch_oddsportal_results_next_scheduled` log after the job completes, success or partial failure
- [ ] Existing `test_fetch_oddsportal_results.py` suite passes; add a test where the scrape returns a record with an id that would trip connection-refresh behaviour

## Notes

- `job_alert_context` (in `odds_core.alerts`) catches exceptions, sends a critical alert, then re-raises. The self-schedule at `fetch_oddsportal_results.py:299` sits *outside* that context (correct), but inside the `main()` function — and `await process_results(sport=sport)` at line 284 is *inside* the context, so any raise there returns from `main()` before the schedule call. The structural fix is to make self-schedule run regardless of the scrape outcome.
- Evidence from CloudWatch 2026-04-20 10:02:34 UTC: `InterfaceError: connection is closed` on `op_live_CRYPAL_NEW_2026-04-11T1400` (a legitimate event, not a ghost — created by migration cc937c8cd362). Duration trajectory 368s → 447s → 472s → 511s across retries.
- Unrelated to #338 / #341 / #342 (OddsPortal frankenstein / Terraform bump / cleanup migration). Pre-existing bug surfaced when the post-deploy smoke test tried to invoke this job.

Acusick1:
## Summary

- Split `process_results()` in `packages/odds-lambda/odds_lambda/jobs/fetch_oddsportal_results.py` into three phases with distinct DB session scopes: read pending events into a detached `PendingEvent` dataclass and close the session, run `run_harvester_historic()` with no open session, open a fresh session for per-record writes with per-record try/except + rollback.
- Moved `main()`'s `schedule_next_execution` call into a `try/finally` outside `job_alert_context` so self-scheduling always runs regardless of whether `process_results()` raised. Scheduling-only failures are now logged and dispatch a critical Discord alert (same `send_critical` pattern used by `fetch_scores` / `fetch_odds` / `fetch_polymarket`) so the EventBridge rule cannot go silently DISABLED.
- Stat counters (`stats.events_updated`, `stats.snapshots_stored`) advance only after `session.commit()` succeeds, so a rolled-back record cannot be counted as persisted.
- Added tests for both the `process_results`-raises-entirely path and the per-record write-failure path, asserting `main()` still reaches `schedule_next_execution`.

## Closes #343

### DIFF
--- a/packages/odds-lambda/odds_lambda/jobs/fetch_oddsportal_results.py
+++ b/packages/odds-lambda/odds_lambda/jobs/fetch_oddsportal_results.py
@@ -254,9 +254,6 @@ async def process_results(
                     away_score=away_score,
                 )
 
-                # Remove from pending list so it can't match again
-                pending_events.remove(event)
-
                 snapshot_stored = False
                 market_data = record.get(market_key, [])
                 if market_data:
@@ -286,11 +283,13 @@ async def process_results(
 
                 await session.commit()
 
-                # Only advance stats after the commit succeeds so rolled-back
-                # records don't inflate success counters.
+                # Only advance batch state after the commit succeeds so rolled-back
+                # records don't inflate success counters or prematurely drop events
+                # from the in-flight dedupe list.
                 stats.events_updated += 1
                 if snapshot_stored:
                     stats.snapshots_stored += 1
+                pending_events.remove(event)
 
             except Exception as e:
                 msg = f"{record.get('home_team', '?')} vs {record.get('away_team', '?')}: {e}"
--- a/packages/odds-lambda/odds_lambda/jobs/fetch_oddsportal_results.py
+++ b/packages/odds-lambda/odds_lambda/jobs/fetch_oddsportal_results.py
@@ -253,11 +253,11 @@ async def process_results(
                     home_score=home_score,
                     away_score=away_score,
                 )
-                stats.events_updated += 1
 
                 # Remove from pending list so it can't match again
                 pending_events.remove(event)
 
+                snapshot_stored = False
                 market_data = record.get(market_key, [])
                 if market_data:
                     match_dt = parse_match_date(record["match_date"])
@@ -282,9 +282,15 @@ async def process_results(
                                 snapshot_time=snapshot_time,
                             )
                             snapshot.api_request_id = API_REQUEST_ID
-                            stats.snapshots_stored += 1
+                            snapshot_stored = True
 
                 await session.commit()
+
+                # Only advance stats after the commit succeeds so rolled-back
+                # records don't inflate success counters.
+                stats.events_updated += 1
+                if snapshot_stored:
+                    stats.snapshots_stored += 1
 
             except Exception as e:
                 msg = f"{record.get('home_team', '?')} vs {record.get('away_team', '?')}: {e}"
@@ -349,6 +355,18 @@ async def main(ctx: JobContext) -> None:
             )
         except Exception as e:
             logger.error("fetch_oddsportal_results_scheduling_failed", error=str(e), exc_info=True)
+            # Escalate to Discord so the failure is visible. Alert dispatch is
+            # best-effort — any error here must not mask an upstream exception
+            # propagating through this ``finally`` block.
+            try:
+                from odds_core.alerts import send_critical
+
+                await send_critical(
+                    f"🚨 fetch-oddsportal-results self-schedule failed "
+                    f"(job={schedule_job_name}): {type(e).__name__}: {e}"
+                )
+            except Exception:
+                logger.warning("fetch_oddsportal_results_scheduling_alert_failed", exc_info=True)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


Acusick1 (commented):
## Review

Addresses #343 cleanly. Three-phase split matches the issue's prescribed approach exactly — session released before the 6–8 min scrape, fresh session for writes, `try/finally` around `schedule_next_execution`. Tests assert the exact failure modes from the issue.

### Strengths

- `PendingEvent` dataclass cleanly detaches data from the session; selecting specific columns avoids hydrating full `Event` objects we can't use after close.
- Per-record `commit`/`rollback` with stats increment *after* commit — the counter semantics are now honest.
- Alert dispatch in the `finally` is wrapped in its own `try/except` so a Discord failure can't mask an upstream exception propagating through `finally` (exactly the subtle trap I'd have flagged otherwise).
- Both test scenarios from success criteria are covered: full raise and per-record failure.

### Minor observations (non-blocking)

- `pending_events.remove(event)` now runs *before* `session.commit()`. If the commit rolls back, the event stays removed from the in-memory list — but since matching is batch-local and the event is still SCHEDULED in DB, next run's `get_pending_events()` picks it up again. No behavioural bug, just noting.
- Per-record commits add DB round trips vs the old single batch commit. Trivial for EPL's ~10 records/day; worth revisiting if this pattern gets reused for a higher-volume league.
- `send_critical` import inside the `except` keeps the hot path lean but means a broken import only surfaces in the failure branch. Given CI exercises both paths, acceptable.

### Nits

- `send_critical` alert string embeds `{type(e).__name__}: {e}` — fine, just confirm Discord handles the `🚨` emoji consistently across the other `send_critical` call sites (this matches the pattern mentioned in the description).
- The docstring on `process_results` is now substantially more useful — keep that style.

Looks good to merge once you've confirmed on prod that the EventBridge rule re-enables after the first post-merge run (success criterion 2 from #343).

Acusick1:
Review follow-up:

**Addressed**
- Observation 1 — `pending_events.remove(event)` now runs after `session.commit()` succeeds, aligning with the same "batch state advances only on commit" invariant already applied to the stat counters. If commit rolls back, the event stays in the in-flight dedupe list (still SCHEDULED in DB, picked up next run). Commit on branch.
- Nit 1 — verified: all existing `send_critical` callers use the 🚨 prefix (`lambda_handler.py:128`, `fetch_scores.py:89`, `fetch_odds.py:150,185`, `fetch_polymarket.py:172`). The new call matches.

**Deferred**
- Observation 2 — per-record commit round trips. EPL volume is ~10 records/day; the scope of #343 explicitly required per-record failure isolation. Worth revisiting if this pattern is reused for a higher-volume league, as you noted.
- Observation 3 — `send_critical` import inside the `except`. Low risk (imports don't rot silently), and moving it to module-top has marginal benefit given the Lambda import surface. Leaving as-is.

Post-merge: will confirm the EventBridge rule returns to ENABLED after the first prod run.

PR #349, opened 2026-04-20
Replace get_sharp_soft_spread with lean find_retail_edges tool

### LINKED ISSUE #348: Redesign get_sharp_soft_spread as a lean edge-screening tool
## Goal

Replace `get_sharp_soft_spread` with a leaner, screening-optimised MCP tool. The current response returns every retail book's per-outcome price verbosely (~54 entries per EPL event) and has demonstrably failed as an edge-discovery API — the agent applied a learned obs-log heuristic to exclude high-hold books wholesale and missed a genuine longer-than-sharp outlier. The new tool pre-computes best/worst retail prices per outcome, dispersion stats, a `z_score` per entry, and a ranked list of retail edges, so the agent gets a signal it cannot dismiss as noise without cause.

This is a **breaking change** — the old tool and its response shape are removed. MCP tools have no external consumers; the only caller is the betting agent, whose prompt we control.

## Motivating example

2026-04-20 EPL session (`logs/agent_runs/soccer_epl_20260420T091836Z_1328844.jsonl`), CRY-WHU brief:

- Sharp (Betfair Exchange) West Ham: +198, implied 33.6%
- **midnite West Ham: +230, implied 30.3%, divergence −3.3%** (longer than sharp)
- Agent brief summary: **"No retail LONGER than sharp"** — literally false.
- Agent body: *"All 17 trusted retail books are SHORTER than sharp on every outcome (Midnite excluded per obs-log — 7.3% overround noise)."*

The agent did **not** fail to see midnite — it saw the entry and deliberately excluded it, citing a learned obs-log rule about high-hold books. A second case in the same run (BUR-MANCIT midnite Man City, −2.77% divergence) was dropped the same way. Both are textbook asymmetric-shading patterns: a high-hold book is long on one outcome *because* it loads hold on the others. The agent's filter is discarding the exact signal the thesis wants to exploit.

Ranking these entries at position 1 of `retail_edges` with a `z_score` of ~−2σ and a tight `dispersion_stddev` makes them much harder to dismiss. It does not fully solve the problem — the prompt must also explicitly reconcile "high book hold" with "long on one outcome" as the pattern, not a contradiction. See Scope.

## Proposed response shape

`per_outcome` is a flat array rather than a dict so totals with multiple `point` values (MLB O/U 7.5/8.5/9.5, future EPL lines) produce the same shape without reshaping.

```jsonc
{
  "event": {...},
  "snapshot_time": "2026-04-20T08:39:31+00:00",
  "sharp_source": "betfair_exchange",
  "per_outcome": [
    {
      "outcome": "West Ham",
      "point": null,
      "sharp_implied_prob": 0.336,
      "sharp_snapshot_time": "2026-04-20T08:39:31+00:00",
      "sharp_age_seconds": 0,
      "best_retail":  {"book": "midnite", "price": 230, "implied_prob": 0.303, "divergence": -0.033, "z_score": -2.2, "market_hold": 0.073},
      "worst_retail": {"book": "10bet",   "price": 170, "implied_prob": 0.370, "divergence":  0.034, "z_score":  2.3, "market_hold": 0.062},
      "n_books": 18,
      "median_divergence": 0.018,
      "dispersion_stddev": 0.015
    }
  ],
  "retail_edges": [
    {"outcome": "West Ham", "point": null, "book": "midnite", "price": 230, "implied_prob": 0.303, "divergence": -0.033, "z_score": -2.2, "market_hold": 0.073}
  ]
}
```

Fields:
- `sharp_implied_prob`, `sharp_snapshot_time`, `sharp_age_seconds`: per-outcome sharp reference with provenance (lookback window preserved).
- `best_retail`: most generous retail price (smallest `implied_prob`, most-negative `divergence`).
- `worst_retail`: least generous retail price. Kept because the obs-log already uses it for asymmetric-hold reads ("Betway shortest on underdogs").
- `n_books`, `median_divergence`, `dispersion_stddev`: noise baseline.
- `z_score`: `(divergence − median_divergence) / dispersion_stddev`. `null` if `n_books < 3` or `stddev == 0`. Cheap discriminator between "typical" and "outlier" that the agent would otherwise have to derive.
- `retail_edges`: flat across all `(outcome, point)` pairs, filtered to `divergence < 0`, sorted ascending (most generous first). Ties break on `z_score` then `book` name for determinism.
- **Full per-book soft list is not returned.** The derived signals replace it.

When sharp is missing for an outcome: `per_outcome` entry present with `sharp_implied_prob: null`, `best_retail`/`worst_retail` populated but no divergence/z_score, no entry in `retail_edges`.

## Scope

- New `find_retail_edges` function in `packages/odds-mcp/odds_mcp/server.py` implementing the shape above.
- **Remove** `get_sharp_soft_spread` and its tests. Hard replacement.
- Rewrite `packages/odds-mcp/agents/common.md` §Edge Types #2 and §Wake-Up Workflow step 5. The "scan the soft array" guidance added in PR #346 becomes obsolete. New language must also explicitly state: **a high market_hold on the book does NOT invalidate a longer-than-sharp entry on one outcome — that is the asymmetric-shading pattern itself. Judge on `z_score` and `dispersion_stddev`, not the book's overall hold.**
- Update `packages/odds-mcp/agents/mlb/base.md` tool-defaults line.
- Update `packages/odds-mcp/agents/epl/base.md` §Edge Types #2 (indirect references to the old scan framing).
- Update `docs/BETTING_AGENT.md` lines 132, 162, 214 for the rename.
- **Local-only (gitignored, not in PR diff, but effective for the agent):** append dated entries to `packages/odds-mcp/agents/epl/observations.md` and `packages/odds-mcp/agents/mlb/observations.md` flagging the graduated "no longer-than-sharp retail on UK books" rule as suspect, since it was built against a tool that suppressed the evidence. The agent reads obs-log at session start and will otherwise continue to cite the graduated rule as strong evidence (6 slates, n~50 fixtures, "high confidence").

## Location

- `packages/odds-mcp/odds_mcp/server.py` — new function, old removed
- `packages/odds-mcp/agents/common.md`, `packages/odds-mcp/agents/mlb/base.md`, `packages/odds-mcp/agents/epl/base.md` — prompt updates
- `docs/BETTING_AGENT.md` — rename references
- `tests/unit/test_mcp_server.py` and `tests/integration/test_mcp_tools.py` — replace `TestGetSharpSoftSpread*` classes with `TestFindRetailEdges`
- Local only: `packages/odds-mcp/agents/{epl,mlb}/observations.md`

## Success criteria

- [ ] New tool callable with `event_id`, `market`, optional `sharp_bookmakers` override, optional `sharp_lookback_hours`; returns the shape above.
- [ ] `get_sharp_soft_spread` is removed from the codebase — no dangling references in code, tests, prompts, or docs.
- [ ] `per_outcome` is a flat array keyed by `(outcome, point)`; h2h/1x2 entries have `point: null`; totals produce one entry per `(outcome, point)`.
- [ ] `best_retail` / `worst_retail` selection correct on h2h, 1x2, and totals markets; tie-breaking deterministic.
- [ ] `retail_edges` contains only entries where `divergence < 0`, sorted by divergence ascending with documented tie-breaks.
- [ ] Empty `retail_edges` when no negative divergence exists.
- [ ] Dispersion stats (`median_divergence`, `dispersion_stddev`) and `z_score` computed correctly; `n_books` matches raw soft count.
- [ ] `market_hold` populated on h2h/1x2, `null` on totals/spreads, `null` for books missing outcomes (existing regression preserved).
- [ ] Pytest class-based tests cover: shape, sign convention, tie-breaking determinism, empty-retail, missing-sharp, multi-line totals, z_score arithmetic on a hand-checked fixture, CRY-WHU-shaped scenario (midnite entry surfaces at rank 1 with expected z_score).
- [ ] Prompt text in `common.md` explicitly states the high-hold-vs-long-on-one-outcome rule.
- [ ] Manual verification: running an EPL wake-up against current data produces a brief that cites `retail_edges` content when a CRY-WHU-shaped scenario exists.
- [ ] Local obs-log entries appended on the agent host so the next session doesn't re-apply the graduated rule.

## Constraints

- Sign convention unchanged: `divergence = retail_implied_prob - sharp_implied_prob`; negative = retail longer than sharp. `z_score` shares the sign (negative for longer-than-sharp outliers).
- Async/SQLModel patterns consistent with the rest of `server.py`; NullPool on Lambda preserved.
- No new dependencies. Reuse `OddsReader.get_sharp_prices`, `extract_odds_from_snapshot`, `per_book_market_holds`, `calculate_implied_probability`.
- Sport-agnostic — h2h, 1x2, totals, spreads all produce the same response shape.

## Decisions (resolved)

- **Tool name:** `find_retail_edges`. Says what it's for; aligns with the primary output key.
- **Totals with multiple point values:** `per_outcome` is a flat array with `(outcome, point)` entries from day one. EPL scraper only has 2.5 today; shape must not need a breaking change when MLB/future lines land.
- **Server-side `market_hold` filter on `retail_edges`:** no. CRY-WHU is the counter-example — the filter would hide the target signal. `market_hold` included per entry so the prompt can weigh tradeability.
- **`z_score` per edge:** yes. Primary discriminator between "noise" and "genuine outlier"; removes arithmetic the agent would otherwise have to do in its head.
- **Drop `worst_retail`:** no, keep. Obs-log already uses it for asymmetric-shading reads. Cheap to include.

## Notes

- Hard replacement. Old tool shape preserved only in git history.
- PR #346's "scan the soft array" guidance (commit e2a1fb5) becomes obsolete the moment this ships — delete it in the same PR.
- The proximate cause of the CRY-WHU miss was a **prompt + obs-log interaction**, not only the tool shape. The redesign helps (ranking + z_score make the signal harder to dismiss), but without the common.md high-hold reconciliation and the obs-log reset, the agent will continue to apply the learned filter to the new tool's output.
- `_DEFAULT_RETAIL_BOOKMAKERS` in `server.py:60` is unrelated (only affects `get_event_features`). `_DEFAULT_SHARP_BOOKMAKERS` duplication with `feature_extraction.py` (flagged in the existing comment) is also out of scope — separate followup.
- Related: #347 (totals betting enablement). This tool's multi-point handling is built in so `retail_edges` becomes the primary discovery signal for O/U edges once #347 ships.

Acusick1:
## Summary

- Hard-replaces `get_sharp_soft_spread` with `find_retail_edges` in the MCP server. The old tool returned ~54 verbose per-book entries per event and had demonstrably failed as an edge-discovery API — the agent was applying a learned obs-log heuristic to exclude high-hold books wholesale, suppressing genuine longer-than-sharp outliers (CRY-WHU midnite West Ham +230 at −3.3%, BUR-MANCIT midnite Man City at −2.77%).
- New response pre-computes `best_retail` / `worst_retail` per `(outcome, point)`, dispersion stats (`n_books`, `median_divergence`, `dispersion_stddev`), a per-entry `z_score`, and a ranked `retail_edges` list so the signal arrives at the agent as a ranked decision-ready object rather than raw data to scan.
- `per_outcome` is a flat array keyed by `(outcome, point)` — future-proofs multi-line totals (MLB O/U 7.5/8.5/9.5) with no reshape.
- Prompts updated: `common.md` §Edge Types #2 and §Wake-Up Workflow step 5 describe `retail_edges` directly and explicitly state that a high `market_hold` on the book does NOT invalidate a longer-than-sharp entry on one outcome — that is the asymmetric-shading pattern itself. Judge on `z_score` and `dispersion_stddev`, not the book's overall hold.
- Sport base files (`epl/base.md`, `mlb/base.md`) and `docs/BETTING_AGENT.md` updated for the rename.
- Tests rewritten (13 unit + 8 integration in `TestFindRetailEdges*`), covering shape, sign convention, cross-bucket sort determinism, empty-retail, zero-retail-books, missing-sharp, market_hold nulling on partial books and on totals/spreads, z_score arithmetic on a hand-checked fixture, a CRY-WHU-shaped fixture asserting the outlier surfaces at `retail_edges[0]`, and multi-line totals producing distinct `(outcome, point)` entries.

## Post-merge action (local, gitignored)

The agent's uncommitted `observations.md` files on the agent host need a dated entry flagging the graduated "no longer-than-sharp retail on UK books" rule (6 slates, n~50 fixtures) as suspect — it was built against a tool that obscured the signal and must be re-tested from scratch against `find_retail_edges` output before re-applying. Entries have been written locally on the dev host but will not appear in this diff.

## Closes #348

### DIFF
--- a/docs/BETTING_AGENT.md
+++ b/docs/BETTING_AGENT.md
@@ -129,7 +129,7 @@ Implemented via the existing `LocalSchedulerBackend` with proximity-aware schedu
 | `save_match_brief` | Persist analysis brief with decision + summary (append-only) |
 | `get_match_brief` | Load full briefs for a single event |
 | `get_slate_briefs` | Latest decision/summary per event for slate triage |
-| `get_sharp_soft_spread` | Dedicated sharp vs soft divergence view |
+| `find_retail_edges` | Ranked retail-vs-sharp divergence screener with per-outcome dispersion stats |
 
 ## Data Model Additions
 
@@ -159,7 +159,7 @@ Build the unified wake-up workflow and iterate interactively.
 
 - Rewrite agent prompts with information-edge thesis and conviction framework
 - Add `match_briefs` table + migration
-- Add `save_match_brief`, `get_match_brief`, `get_sharp_soft_spread` MCP tools
+- Add `save_match_brief`, `get_match_brief`, `find_retail_edges` MCP tools
 - Run wake-up workflow interactively for 2-3 matchdays
 - Iterate on prompt: which research patterns surface actionable info? Where does the agent waste time?
 - Identify reliable lineup/team news data sources (API-Football free tier insufficient — see `docs/AGENT_DATA_SOURCES.md`)
@@ -211,7 +211,7 @@ The agent must demonstrate it adds value beyond a simple mechanical rule. Track 
 
 ## Role of the XGBoost Model
 
-The CLV prediction model (R² = 3-6%, backtest ROI +3.4% at p = 0.26) is demoted to a supplementary signal. Its strongest feature (`retail_sharp_diff`) is the sharp-soft spread, which the agent can observe directly via `get_sharp_soft_spread`.
+The CLV prediction model (R² = 3-6%, backtest ROI +3.4% at p = 0.26) is demoted to a supplementary signal. Its strongest feature (`retail_sharp_diff`) is the sharp-soft spread, which the agent can observe directly via `find_retail_edges`.
 
 `get_predictions` remains available as a sanity check. The agent prompt says: "Model predictions are weakly predictive. Do not bet based on model output alone."
 
--- a/packages/odds-mcp/agents/common.md
+++ b/packages/odds-mcp/agents/common.md
@@ -26,7 +26,9 @@ You are not limited to one edge type. Explore all four. Over time we will measur
 The market has not yet absorbed a piece of news. Edge window for public information is minutes to hours, not days. Key player ruled out / surprise starter / late tactical change / late scratch are canonical examples.
 
 ### 2. Retail / cross-venue dispersion
-A retail book offers odds longer than sharp on some outcome, or book-to-book dispersion is wider than normal. `get_sharp_soft_spread` returns every non-sharp book in the latest snapshot — typically 15–20 UK/EU books for EPL, with per-outcome `divergence` (soft minus sharp implied probability) and per-book `market_hold`. A **negative divergence** means the retail book is pricing that outcome *longer* than sharp — a candidate edge. Common causes: stale pricing, a catalyst still propagating, or book-specific shading asymmetry (a book loading hold on one side to manage liability while leaving the other side near sharp).
+A retail book offers odds longer than sharp on some outcome, or book-to-book dispersion is wider than normal. `find_retail_edges` returns a pre-ranked `retail_edges` list — every `(outcome, point)` entry where at least one retail book's implied probability is below the sharp implied probability (**negative divergence** = retail priced *longer* than sharp). Each entry carries `z_score` against the per-outcome retail pack and `market_hold` for tradeability context. Common causes: stale pricing, a catalyst still propagating, or book-specific shading asymmetry (a book loading hold on one side to manage liability while leaving the other side near sharp).
+
+A high `market_hold` on the book does **NOT** invalidate a longer-than-sharp entry on one outcome — that is the asymmetric-shading pattern itself. Judge on `z_score` magnitude and `dispersion_stddev`, not the book's overall hold.
 
 ### 3. Structural biases
 Bookmaker pricing mechanics create systematic mispricings unrelated to match-specific information — public money loading on popular teams, accumulator distortion on popular legs, televised-match liability shading, mid-season calibration on promoted/relegated or newly-prominent teams. The question is not whether these happen (they do) but whether the shading in any given match exceeds what sharp already accounts for.
@@ -120,7 +122,7 @@ For any fixture without a prior brief, write your own probability estimate from 
 ### 5. Research
 Concurrent where possible (see Parallelism):
 
-- Check sharp-soft spread via `get_sharp_soft_spread`. The response returns **every non-sharp book in the latest snapshot** (not a pre-filtered short list). Scan the `soft` array per outcome for **negative `divergence` values** — these are retail books pricing longer than sharp. Flag the most-divergent book per outcome in your brief even when it is not a household retail name. Weigh `market_hold` when judging tradeability: very high hold (>7%) often indicates a rec book with unsustainable listings, but a single-outcome longer-than-sharp price is still value if the book will take the bet.
+- Call `find_retail_edges` and inspect the `retail_edges` array. An empty array means no retail book is pricing any outcome longer than sharp — move on. A rank-1 entry with `z_score ≤ −1.5` against a tight `dispersion_stddev` is the signal to investigate — the pack agrees the price is out, and the outlier is materially out, not rounding noise. A negative divergence is necessary but not sufficient: size the edge against the retail odds and confirm EV flips at the offered price, not just at the implied probability.
 - Refresh scrape if odds data is stale (`refresh_scrape`).
 - Web search for team news, injuries, lineups, tactical context.
 - Compare current sharp price against any previous brief's sharp price and against your pre-market read.
--- a/packages/odds-mcp/agents/epl/base.md
+++ b/packages/odds-mcp/agents/epl/base.md
@@ -28,7 +28,7 @@ Concrete examples that fit each `common.md` edge category in an EPL context:
 
 **1. Information gaps** — key player ruled out minutes before lineup release; unexpected starter or bench; manager quotes signalling rotation or tactical change; late goalkeeper change (high-impact, often underweighted).
 
-**2. Retail / cross-venue dispersion** — the OP UK-retail book set is homogeneous on 1x2. Genuine longer-than-sharp observations are rare; when they appear, usually indicate stale retail pricing or a catalyst still propagating.
+**2. Retail / cross-venue dispersion** — the OP UK-retail book set is homogeneous on 1x2. Genuine longer-than-sharp observations are rare; when `find_retail_edges` returns a rank-1 entry, it usually indicates stale retail pricing, a catalyst still propagating, or asymmetric liability shading on a high-hold book. Do not discard high-hold outliers — a book can load hold on two outcomes while leaving the third long, and that is exactly the pattern to investigate.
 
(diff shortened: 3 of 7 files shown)


Acusick1 (commented):
## Review

Read PR body, issue #348, full diff, and ran the new tests (`TestFindRetailEdges` — 13 unit + 8 integration) — all 21 pass. CI green. Ruff clean. No dangling `get_sharp_soft_spread` references anywhere (code/tests/prompts/docs).

(Can't formally approve my own PR — posting as comment; treat as approve-with-nits.)

### What works well

- Response shape is decision-ready: pre-ranked `retail_edges`, per-(outcome, point) dispersion baseline, per-row `z_score`. This is the right abstraction for the agent's failure mode on CRY-WHU — the old tool dumped raw rows and the agent's learned high-hold filter silently discarded the signal.
- `(outcome, point)` flat array generalises to MLB totals without a shape change. The multi-line totals integration test locks that in.
- `common.md` reconciles the asymmetric-shading / high-hold contradiction explicitly — this is load-bearing for the behavioural fix that the issue highlighted. Without it, the tool redesign alone wouldn't have moved the agent.
- Sharp-lookback resolution preserved and tested (`TestFindRetailEdgesSharpLookback`), including the null-sharp-within-window edge case.
- `market_hold` regression from the prior PR preserved — null on totals/spreads and on partial books — with both paths covered. The `test_partial_book_market_hold_none` tightening (partial_book at +120 so it wins `best_retail`) is the right way to exercise the null propagating into the surfaced row.
- Second commit's rename `sharp_source → sharp_bookmakers` is a genuine correctness fix: under hybrid fallback the configured list is plural and per-outcome resolution is visible via `sharp_snapshot_time` / `sharp_age_seconds` anyway. Good catch.

### Minor — not blocking

1. `server.py:1113` — the retail_edges sort key's final tuple `(e[\"point\"] is None, e[\"point\"] if e[\"point\"] is not None else 0.0)` is equivalent to `math.inf if e[\"point\"] is None else e[\"point\"]`, matching `_bucket_sort_key`'s simpler form at `server.py:1008`. Consistency would read better; purely stylistic.
2. `tests/integration/test_mcp_tools.py` CRY-WHU fixture: `assert edges[0][\"market_hold\"] > 0.05` is a loose floor (the actual midnite hold from the fixture prices is ~18%). Fine as a sanity check; flagging only so future hand-edits of the prices don't silently drift past the threshold without noticing.
3. `per_book_market_holds` is called with `required_outcomes = {o.outcome_name for o in odds}` i.e. the union across sharp + retail. Correct for h2h/1x2 where the outcome set is shared, but worth noting: if sharp ever exposes an outcome retail doesn't carry (or vice versa), the min-coverage requirement would silently drop books. Not a concern with current data, just a sharp corner to remember if we ever mix source markets.

### Scope note

Noted the PR body's mention of local `observations.md` edits on the agent host (not in diff, gitignored). That's the right place for them, and the graduated-rule reset is important — without it the agent will re-apply the filter to the new tool's output and we'll be right back at CRY-WHU.

PR #352, opened 2026-04-21
Add ESPN fixture ingestion job and get_team_context MCP tool

Acusick1:
## Summary

- Extracted `EspnFixtureFetcher` from `scripts/ingest_espn_fixtures.py` into a reusable async class in `packages/odds-lambda/odds_lambda/espn_fixture_fetcher.py` (httpx-based, retry with backoff, dedup, `current_season` helper).
- Refactored the script to delegate to the fetcher while retaining multi-season backfill + CSV capability.
- New scheduled job `fetch-espn-fixtures-epl` at `packages/odds-lambda/odds_lambda/jobs/fetch_espn_fixtures.py`, registered in `scheduling/jobs.py` and in Terraform (`eventbridge.tf`) as a fixed daily cron at 06:00 UTC. EPL-only. Cadence owned by Terraform cron (no self-scheduling — matches `daily-digest` / `score-predictions`).
- New `get_team_context` MCP tool in `packages/odds-mcp/odds_mcp/server.py` returning `last_results` (PL-only, Final), `upcoming_fixtures` (all competitions, with `days_until_ko`), and derived PL `standings` (points → GD → GF tiebreak) for any normalised EPL team. `as_of` cutoff honoured; mid-match `In Progress` rows skipped from both sides.
- Tests: 9 fetcher tests (mocked httpx), 15 `get_team_context` tests covering standings derivation, tiebreaks, PL-only filter, `as_of` no-look-ahead, mid-match skip, and alias normalisation.

## Closes #351

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #353, opened 2026-04-21
Fix ESPN fixture status filter and add upcoming-fixture fetch

Acusick1:
## Summary

Post-merge smoke testing against live ESPN data revealed `get_team_context` (shipped in #352) returns empty `last_results`, `upcoming_fixtures`, and `standings` for every team.

**Root causes**
1. `_ESPN_STATUS_FINAL = "Final"` never matches — ESPN actually emits `"Full Time"`, `"Final Score - After Penalties"`, `"Final Score - After Extra Time"`.
2. ESPN's `teams/{id}/schedule` endpoint returns only completed matches, so there was no upcoming-fixture data source.

## Changes

- **Schema**: added nullable `state` column (`"pre"` / `"in"` / `"post"`) to `EspnFixture`, populated from ESPN's canonical `status.type.state`. Alembic migration `beaf16386050`.
- **Fetcher**: `fetch_team_schedule` now captures `state`. New `fetch_upcoming(days_ahead=30)` hits the scoreboard endpoint per league slug, synthesising two team-anchored records per match.
- **Job**: runs both past (team-schedule) + upcoming (scoreboard) fetches and upserts the combined set. Scoreboard wins on `(date, team, opponent)` conflicts so just-kicked-off matches get the freshest `state`.
- **MCP tool**: replaces the status-string constants with state-based helpers `_is_final` / `_is_in_progress`, with a fallback to a named set of final-status strings for legacy null-state rows.
- **Tests**: updated unit-test mocks to match real ESPN shape; added null-state fallback test; added scoreboard-mock unit test; added `tests/integration/test_espn_live_api.py` that hits the live API (marked `pytest.mark.integration`) so future ESPN shape changes get caught pre-merge.
- Narrowed broad exception handlers in fetcher loops; dropped unused `_ESPN_STATE_PRE` constant.

## Verification

Local smoke test against live ESPN + local DB:
- `post`: 867 rows, `pre`: 110 rows, no NULL state
- `get_team_context(team="Arsenal")` returns populated `last_results` (3), `upcoming_fixtures` (3), standings `position=1, points=70, played=33, table_size=20`

### DIFF
--- a/.github/workflows/live-api.yml
+++ b/.github/workflows/live-api.yml
@@ -1,14 +1,13 @@
-name: Test
+name: Live API Tests
 
 on:
-  push:
-    branches-ignore: [main]
-  workflow_call:
+  schedule:
+    - cron: "0 6 * * 1"
   workflow_dispatch:
 
 jobs:
-  test:
-    name: Run Unit and Integration Tests
+  live-api:
+    name: Run live-API integration tests
     runs-on: ubuntu-latest
     environment: development
 
@@ -29,23 +28,17 @@ jobs:
       - name: Install dependencies
         run: uv sync
 
-      - name: Run unit tests
+      - name: Run live-API tests
         env:
           DATABASE_URL: ${{ secrets.DATABASE_URL }}
           ODDS_API_KEY: ${{ secrets.ODDS_API_KEY }}
-        run: uv run pytest tests/unit -v
-
-      - name: Run integration tests
-        env:
-          DATABASE_URL: ${{ secrets.DATABASE_URL }}
-          ODDS_API_KEY: ${{ secrets.ODDS_API_KEY }}
-        run: uv run pytest tests/integration -v -m "not live_api"
+        run: uv run pytest -v -m live_api
 
       - name: Upload logs on failure
         if: failure()
         uses: actions/upload-artifact@v4
         with:
-          name: test-logs
+          name: live-api-logs
           path: |
             *.log
             *.txt
--- a/migrations/versions/beaf16386050_add_state_column_to_espn_fixtures.py
+++ b/migrations/versions/beaf16386050_add_state_column_to_espn_fixtures.py
@@ -1,0 +1,28 @@
+"""add state column to espn_fixtures
+
+Revision ID: beaf16386050
+Revises: 8a76c9607193
+Create Date: 2026-04-21 13:36:34.478454
+
+"""
+
+import sqlalchemy as sa
+import sqlmodel
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "beaf16386050"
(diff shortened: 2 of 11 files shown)


Acusick1 (commented):
Looks solid — well-scoped fix for the two root causes the PR identifies, and the MCP refactor is cleaner than the status-string matching it replaces. CI green, schema change is forward-compatible (nullable `state` + fallback set covers legacy rows).

**A few non-blocking observations**

1. `except (httpx.HTTPError, httpx.TimeoutException)` — `TimeoutException` is already a subclass of `HTTPError` (via `TransportError → RequestError → HTTPError`), so the tuple is redundant. Just `httpx.HTTPError` is equivalent. Happens in `_fetch_json`, `fetch_scoreboard`, and `fetch_season`.

2. `tests/integration/test_espn_live_api.py` uses `pytestmark = pytest.mark.integration` but the marker isn't registered in `pyproject.toml` (`[tool.pytest.ini_options] markers = [...]`). It'll emit `PytestUnknownMarkWarning`. Also worth noting: CI's `Run integration tests` step runs everything under `tests/integration/` unconditionally — the mark is advisory, not a gate. Fine given `pytest.skip` on empty responses, but if ESPN itself 5xxs the test will fail (not skip). Catching `httpx.HTTPError → pytest.skip` would make it more robust against transient upstream issues.

3. Job-level dedup key `(date.isoformat(), team, opponent)` diverges from the DB unique constraint `(date, team, competition)`. Equivalent in practice (a team can't play two matches at the same timestamp), but the mismatch is a minor footgun if the constraint ever changes. Same pattern already in `fetch_season`/`fetch_upcoming` internal dedup, so it's at least consistent.

4. In `fetch_scoreboard`, `label = season_label(current_season(dt))` is computed per-event — good, correctly handles the Jul→Aug season boundary when `fetch_upcoming`'s 30-day window crosses August 1.

5. `_is_final` vs the original `row.date < resolved_as_of` check: the new logic is slightly more conservative — a `state="post"` row with `date >= as_of` (inconsistent but possible) now gets dropped, whereas the old code would've added it to upcoming. This is arguably the right behavior; just flagging it's a subtle behavior change beyond the bug fix.

**Verified**
- Migration `beaf16386050 → 8a76c9607193` chains correctly to the latest head.
- Dedup iteration order (`*upcoming_records, *past_records`) matches the comment and produces the documented "scoreboard wins" semantics.
- `_ESPN_FINAL_STATUS_FALLBACK` is exact-match only — good call; a `startswith("Final")` check would false-positive on "Final Score" during live ESPN response evolution.
- Integration test's `_EXPECTED_STATES = {"pre", "in", "post"}` is the right schema-drift tripwire.

Nothing blocking — ship when ready.

PR #357, opened 2026-04-21
Move scoring + fixed-schedule jobs local; add odds_agent role

### LINKED ISSUE #355: Move prediction scoring and fixed-schedule jobs local; point agent at odds DB
## Goal

The betting agent reads predictions from whatever DB `DATABASE_URL` resolves to (default `odds_test`), but scoring runs in Lambda against prod Neon — so the agent sees stale/empty Predictions. Move scoring and fixed-schedule jobs local so the agent, scraper, and scorer all share one rolling DB, and eliminate the cross-environment data gap.

## Scope

1. **Point agent at local `odds`.** Change `--db` default from `"odds_test"` to `"odds"` in `packages/odds-cli/odds_cli/commands/agent.py:21-25`. Drop `snapshot_to_dev.sh` guidance from the docstring at `27-34`.
2. **Inline `score_events` after `fetch_oddsportal`.** In `packages/odds-lambda/odds_lambda/jobs/fetch_oddsportal.py::main`, after the existing `job_alert_context` block, call `await score_events(sport=ctx.sport)` wrapped in its own try/except so a scoring bug doesn't surface as a scrape alert. Update the stale docstring at `packages/odds-lambda/odds_lambda/jobs/score_predictions.py:8` if needed.
3. **Add CronTrigger support to `LocalSchedulerBackend`.**
   - Add `_JOB_CRON_MAP: dict[str, str]` in `packages/odds-lambda/odds_lambda/scheduling/jobs.py`, mirroring terraform's `fixed_schedule_expressions`:
     - `\"daily-digest\": \"0 8 * * *\"`
     - `\"fetch-espn-fixtures\": \"0 6 * * *\"`
   - Add `schedule_cron(job_name, cron_expr, payload)` to `LocalSchedulerBackend` using `scheduler.add_schedule(job_func, trigger=CronTrigger.from_crontab(...), conflict_policy=ConflictPolicy.replace)`.
   - In `packages/odds-cli/odds_cli/commands/scheduler.py:105-135`, branch the bootstrap loop: jobs with a cron entry go through `schedule_cron`; others use the existing `bootstrap_fn` path.
4. **Promote `bootstrap_jobs` default.** Change default in `packages/odds-core/odds_core/config.py:88-91` from `[\"agent-run\"]` to `[\"agent-run\", \"fetch-oddsportal\", \"fetch-oddsportal-results\", \"daily-digest\", \"fetch-espn-fixtures\"]`. Update the assertion in `tests/unit/test_config.py:39`.
5. **Remove Lambda schedules for jobs that are now local.** In `deployment/aws/terraform/eventbridge.tf`, drop `score-predictions`, `daily-digest`, and `fetch-espn-fixtures` from `fixed_schedule_expressions` (138-142) and `sport_fixed_scheduler_rules` (147-164). Keep `fetch-odds` (deprioritized, out of scope). Do **not** delete the job modules themselves. Requires `terraform apply` after merge.
6. **Update `.env.example`.** Add `MODEL_NAME`, `MODEL_BUCKET`, `SCHEDULER_BACKEND=local`. Local AWS creds with S3 read on the model bucket are required (plain `boto3` in `model_loader.py:22`).
7. **Add read-only Postgres role for the agent (defense-in-depth).** Create role `odds_agent` with `SELECT` on all tables + `INSERT/UPDATE` on `paper_bets`, `match_briefs`, `agent_wakeups` only. Agent process connects as this role. Closes the Bash escape hatch that the MCP tool surface doesn't cover.

## Location

- `packages/odds-cli/odds_cli/commands/{agent.py,scheduler.py}`
- `packages/odds-lambda/odds_lambda/jobs/{fetch_oddsportal.py,score_predictions.py}`
- `packages/odds-lambda/odds_lambda/scheduling/{jobs.py,backends/local.py}`
- `packages/odds-core/odds_core/config.py`
- `deployment/aws/terraform/eventbridge.tf`
- `.env.example`
- `tests/unit/test_config.py`

## Implementation approach

- **Why local scoring is safe:** `score_predictions._extract_features` (`packages/odds-lambda/odds_lambda/jobs/score_predictions.py:80-131`) is deterministic — uses only snapshot odds (frozen `raw_data`) and `event.commence_time - snapshot.snapshot_time`. No `datetime.now()` in the feature path, no look-ahead risk.
- **Why not sync via `snapshot_to_dev.sh`:** it uses `pg_dump | pg_restore --clean --if-exists`, which would wipe the agent's `PaperBet` and `MatchBrief` rows on every sync.
- **Why `CronTrigger` and not self-scheduling:** fixed-schedule jobs (daily-digest, espn-fixtures) have no dynamic fire-time logic, so APScheduler's native `CronTrigger` is sufficient. Self-scheduling (`DateTrigger` + reschedule at end of `main`) remains correct for jobs that compute fire time from game proximity.
- **`score_events` failure modes are already graceful** — returns empty stats dict on missing model, missing config, or sport mismatch (verified in `tests/unit/test_score_predictions.py:206-279`). Inline wrapper just needs to catch any unexpected exception.
- **`copy_from_production.py` is unaffected** — its `_copy_predictions` path uses ON CONFLICT DO NOTHING on `(event_id, snapshot_id, model_name)`, so local predictions always win and prod-only predictions fill gaps. No contamination risk under this change.

## Constraints

- `LocalSchedulerBackend` uses `SQLAlchemyDataStore` + `AsyncpgEventBroker`; new cron schedules are picked up by the running scheduler via LISTEN/NOTIFY.
- Do not delete `packages/odds-lambda/odds_lambda/jobs/{score_predictions,daily_digest,fetch_espn_fixtures}.py` — only their Lambda triggers go away.
- `terraform apply` must run after merge, otherwise EventBridge keeps invoking removed jobs against the scheduler Lambda.

## Success criteria

- [ ] `odds agent run --sport soccer_epl` defaults to local `odds` DB; `get_predictions` returns fresh rows after a `fetch-oddsportal` run
- [ ] `odds scheduler start` bootstraps `fetch-oddsportal`, `fetch-oddsportal-results`, `daily-digest`, `fetch-espn-fixtures`, `agent-run` with no env-var override
- [ ] `daily-digest` and `fetch-espn-fixtures` fire on their cron schedules locally (08:00 and 06:00 UTC)
- [ ] `score_events` runs at the end of each `fetch_oddsportal` scrape; scoring failure does not alert as scrape failure
- [ ] `terraform plan` shows only removals for `score-predictions`, `daily-digest`, `fetch-espn-fixtures` rules; `fetch-odds` unchanged
- [ ] Agent DB connection uses `odds_agent` role; `DROP`/`DELETE` against scraped tables denied
- [ ] `test_config.py::test_settings_defaults` updated and passing
- [ ] `score_predictions.py:8` docstring matches actual behaviour

## Notes

- Predictions consumers verified: MCP `get_predictions`, `daily_digest`, `copy_from_production`, `score_predictions` itself, tests. No analytics or dashboard readers.
- `fetch-odds` Lambda removal is a separate concern tracked elsewhere.
- Neon continues to receive Lambda scrapes with no Predictions — becomes a cold scrape mirror post-change.

Acusick1:
## Summary

- **Agent CLI points at local `odds` DB** (default changed from `odds_test`) so the agent reads predictions from the same database the local scraper and scorer write to.
- **Scoring runs inline at the end of `fetch_oddsportal`** (wrapped in its own try/except so scoring bugs don't surface as scrape alerts). `score-predictions` is no longer a standalone scheduled job.
- **`LocalSchedulerBackend` gains `schedule_cron`.** New `_JOB_CRON_MAP: dict[str, tuple[str, tuple[SportKey, ...] | None]]` drives cron bootstrap; `daily-digest` and `fetch-espn-fixtures` are both gated to `soccer_epl` to avoid daily-critical Discord alerts for unsupported sports (MLB).
- **`bootstrap_jobs` default** promoted to `[agent-run, fetch-oddsportal, fetch-oddsportal-results, daily-digest, fetch-espn-fixtures]` so `odds scheduler start` covers the full local workflow without env overrides.
- **EventBridge no longer schedules** `score-predictions`, `daily-digest`, or `fetch-espn-fixtures`. `fetch-odds` unchanged. `terraform apply` required after merge — otherwise EventBridge keeps invoking removed jobs.
- **`odds_agent` Postgres role** added via migration `f1a2b3c4d5e6`: `SELECT` on all tables + `INSERT/UPDATE` on `paper_trades`, `match_briefs`, `agent_wakeups`. Role is `WITH LOGIN` but no password — operator sets credentials out-of-band. Agent subprocess reads `AGENT_DATABASE_URL` override; falls back to `DATABASE_URL` with a warning log if unset.
- **`heartbeat_expectations`** drops the now-stale `score-predictions-epl` key so health monitor no longer alerts for a job that was renamed to an inline call.
- **`.env.example`** adds `MODEL_NAME`, `MODEL_BUCKET`, `AGENT_DATABASE_URL`, and refreshes the `BOOTSTRAP_JOBS` comment.

## Follow-ups

- Operator must `terraform apply` after merge to drop the now-orphan EventBridge rules.
- Operator must provision the `odds_agent` role password and set `AGENT_DATABASE_URL` to realize the defense-in-depth control; otherwise the subprocess runs with the scraper's write-capable DSN (logged as a warning).
- Doc drift flagged but not addressed in this PR: `docs/ARCHITECTURE.md`, `docs/DEPLOYMENT.md`, and `docs/DEBUGGING_SCHEDULER.md` still describe these jobs as EventBridge-hosted. Worth a follow-up pass.

## Closes #355

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,8 +10,13 @@ SCHEDULER_BACKEND=local
 # How many days ahead to check for games (default: 7)
 SCHEDULER_LOOKAHEAD_DAYS=7
 
-# Jobs to bootstrap on scheduler start (JSON list, default: ["agent-run"])
-# SCHEDULER_BOOTSTRAP_JOBS=fetch-odds,fetch-oddsportal,agent-run
+# Jobs to bootstrap on scheduler start (JSON list). Default includes the
+# event-driven scraper jobs (fetch-oddsportal, fetch-oddsportal-results) and
+# the fixed-schedule jobs (daily-digest, fetch-espn-fixtures) that now run
+# on the local scheduler instead of EventBridge.
+# Default: ["agent-run", "fetch-oddsportal", "fetch-oddsportal-results",
+#           "daily-digest", "fetch-espn-fixtures"]
+# SCHEDULER_BOOTSTRAP_JOBS=fetch-oddsportal,fetch-oddsportal-results,daily-digest,fetch-espn-fixtures,agent-run
 
 # =============================================================================
 # AWS CONFIGURATION (only needed if SCHEDULER_BACKEND=aws)
@@ -52,6 +57,16 @@ PROD_DATABASE_URL=postgresql+asyncpg://user:password@prod-host:5432/odds_prod?ss
 
 # Connection pool size (default: 5)
 DATABASE_POOL_SIZE=5
+
+# Agent subprocess DSN (optional, defense-in-depth).
+# When set, the ``agent-run`` job overrides DATABASE_URL in the agent
+# subprocess with this value so the agent connects as the read-mostly
+# ``odds_agent`` role (created by migration f1a2b3c4d5e6). The role has
+# SELECT on all tables and INSERT/UPDATE only on paper_trades,
+# match_briefs, and agent_wakeups. When unset, the agent inherits the
+# parent's DATABASE_URL (security control disabled) and a warning is
+# logged — functional, but the agent can write to any table.
+# AGENT_DATABASE_URL=postgresql+asyncpg://odds_agent:password@host:5432/odds?ssl=require
 
 # =============================================================================
 # THE ODDS API CONFIGURATION
--- a/.env.example
+++ b/.env.example
@@ -24,6 +24,15 @@ AWS_REGION=us-east-1
 AWS_LAMBDA_ARN=
 
 # =============================================================================
+# MODEL LOADING (required for local score-predictions)
+# =============================================================================
+# Even with SCHEDULER_BACKEND=local, model artifacts are pulled from S3 via
+# plain boto3 (packages/odds-lambda/odds_lambda/model_loader.py). Local AWS
+# credentials with S3 read permission on MODEL_BUCKET are required.
+MODEL_NAME=epl-clv-home
+MODEL_BUCKET=
+
+# =============================================================================
 # DATABASE CONFIGURATION
 # =============================================================================
 # Primary database URL (used by application)
--- a/deployment/aws/terraform/eventbridge.tf
+++ b/deployment/aws/terraform/eventbridge.tf
@@ -135,33 +135,15 @@ locals {
 
   # Schedule expressions for fixed-schedule jobs (map lookup prevents
   # a new job silently inheriting the wrong schedule via a ternary fallback).
-  fixed_schedule_expressions = {
-    "daily-digest"        = "cron(0 8 * * ? *)"
-    "score-predictions"   = "cron(30 * * * ? *)"
-    "fetch-espn-fixtures" = "cron(0 6 * * ? *)"
-  }
+  # score-predictions, daily-digest, and fetch-espn-fixtures now run in the
+  # local scheduler so the agent, scraper, and scorer share one DB. Only
+  # jobs that still fire from EventBridge belong here.
+  fixed_schedule_expressions = {}
 
-  # Per-sport fixed-schedule jobs. fetch-espn-fixtures is EPL-only today —
-  # gated on sport_suffix to avoid scheduling it for sports without ESPN fixture
-  # coverage in this pipeline.
-  sport_fixed_scheduler_rules = flatten([
-    for sport_suffix, cfg in local.sport_configs : concat(
-      [
-        for job in ["daily-digest", "score-predictions"] : {
-          key       = "${job}-${sport_suffix}"
-          job       = job
-          sport_key = cfg.sport_key
-        }
-      ],
-      sport_suffix == "epl" ? [
-        {
-          key       = "fetch-espn-fixtures-${sport_suffix}"
-          job       = "fetch-espn-fixtures"
-          sport_key = cfg.sport_key
-        }
-      ] : []
-    )
-  ])
+  # Per-sport fixed-schedule jobs. Empty until a Lambda-hosted fixed-schedule
+  # job is reintroduced; kept as a list comprehension so adding one back is
+  # a single-line change.
+  sport_fixed_scheduler_rules = []
 
   fixed_scheduler_rules_map = { for r in local.sport_fixed_scheduler_rules : r.key => r }
 }
--- a/deployment/aws/terraform/eventbridge.tf
+++ b/deployment/aws/terraform/eventbridge.tf
@@ -260,7 +260,7 @@ output "dynamic_rules" {
 
 output "scheduler_jobs" {
   description = "Job names routed to the scheduler Lambda (CSV for scripts)"
-  value       = join(",", concat(local.self_scheduling_scheduler_jobs, keys(local.fixed_scheduler_rules_map)))
+  value       = join(",", local.self_scheduling_scheduler_jobs)
 }
 
 output "scraper_jobs" {
--- a/deployment/aws/terraform/eventbridge.tf
+++ b/deployment/aws/terraform/eventbridge.tf
@@ -36,8 +36,11 @@ resource "aws_lambda_permission" "allow_eventbridge_backfill_polymarket" {
   ]
 }
 
-# Per-sport fixed-schedule rules (daily-digest and score-predictions).
-# Generated from sport_configs so each sport gets independent rules.
+# Per-sport fixed-schedule rules. Currently empty — score-predictions,
+# daily-digest, and fetch-espn-fixtures now run on the local scheduler
+# (see `fixed_schedule_expressions` below). The resources are retained so
+# adding a new Lambda-hosted fixed-schedule job is a single-line change
+# (add the job to `fixed_schedule_expressions` + `sport_fixed_scheduler_rules`).
 resource "aws_cloudwatch_event_rule" "fixed_scheduler" {
   for_each = local.fixed_scheduler_rules_map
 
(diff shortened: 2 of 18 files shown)


Acusick1 (commented):
## Review — `/review-pr 357`

**Cannot approve**: both commits are co-authored by Claude, so this is a comment-only review. CI is green.

Overall the change matches the issue scope and the design is sound — scoring inlined with its own try/except outside the scrape alert context, local cron support via APScheduler, sport-gated `_JOB_CRON_MAP`, and a fail-closed `odds_agent` role.

### Strengths
- **Scope discipline**: scoring failure isolated from scrape alerts (`fetch_oddsportal.py:349–363`), confirmed to sit *outside* the `job_alert_context` block.
- **Per-sport allowlist in `_JOB_CRON_MAP`** (`scheduling/jobs.py`) correctly prevents `fetch-espn-fixtures` from raising for MLB every morning — the follow-up commit caught a real behavioural bug.
- **`heartbeat_expectations` drop of `score-predictions-epl`** is necessary and easy to miss; good that it's in.
- **Migration is conservative**: idempotent DO blocks, no-password role fails closed, downgrade hazard documented with `pg_terminate_backend` recipe, `depends_on = beaf16386050` matches current head.
- **`_build_agent_subprocess_env` fallback**: `AGENT_DATABASE_URL` unset → inherits parent DSN with a warning instead of hard-failing, preserving install ergonomics.
- **Typing improvement** on `_write_cron_schedule`/`_add_cron_schedule` (`Callable[[JobContext], Awaitable[None]]`) with existing `_write_schedule` left alone — correct out-of-scope call.
- **Module-level assert** on `_JOB_CRON_MAP ⊆ _JOB_MODULE_MAP` catches typos at import time.

### Concerns

1. **Changed cron expression is sticky across restarts.** The scheduler CLI skips bootstrap when `get_job_status(...).next_run_time` is non-null (`scheduler.py:143–152, 154–167`). Combined with `CronTrigger`'s persistence in the APScheduler data store, if you later edit `_JOB_CRON_MAP` (e.g. move `daily-digest` from `0 8 * * *` to `0 9 * * *`) the stored schedule wins and the new expression never applies. `ConflictPolicy.replace` on `add_schedule` would handle it, but the CLI short-circuits before reaching it. Either drop the skip for cron jobs (the `replace` policy already makes it idempotent), or compare the stored trigger's crontab to the desired expression and re-write on mismatch. Worth at least a comment near the skip check.

2. **Bootstrap loop duplication.** `scheduler.py` now has two parallel branches (cron vs. self-scheduling), each with its own per-sport expansion, existing-check, skip-message, try/except, and success/failure print. The shapes are close enough that a small helper `_bootstrap_for_sport(name, compound, fn)` would cut ~40 lines and make the divergence explicit. Not blocking.

3. **No tests for the new behaviour.** `test_config.py` updates the bootstrap default, but:
   - no test exercises `get_job_cron` / `get_job_cron_sports` (in particular the allowlist gating that was the whole point of the second commit),
   - no test covers `LocalSchedulerBackend.schedule_cron` idempotency or the CronTrigger vs DateTrigger coexistence,
   - no test covers `_build_agent_subprocess_env` overriding vs inheriting `DATABASE_URL`.
   Worth at least a unit test for the first two — they're pure-function and easy.

4. **MCP write surface vs. role grants.** Migration grants `INSERT, UPDATE` on `paper_trades`, `match_briefs`, `agent_wakeups`. Worth double-checking that no MCP tool path issues a `DELETE` on these (e.g. if `save_match_brief` ever supersedes rather than upserts, or if `settle_bets` deletes on a rollback). If any does, it fails silently under the new role. A quick grep on the MCP server would confirm.

5. **Doc drift is acknowledged but not fixed.** `docs/ARCHITECTURE.md:148` still lists `daily-digest` as "Fixed: 08:00 UTC daily" under the EventBridge-hosted jobs table. The PR body calls this out as a follow-up — fine, but please open a tracking issue so it doesn't get lost.

6. **`MODEL_NAME=epl-clv-home` hardcoded in `.env.example`.** Verify this is the currently published model name before merge — `.env.example` defaults flow through to every fresh install. If model naming changes regularly, consider leaving it empty with a comment pointing at `odds model list`.

### Nits
- `score_predictions.py` docstring still says "Called inline at the end of fetch-oddsportal" at line 8 — accurate, but the retained `main()` / manual-invocation story is only in the module-level addition at line 9–11. Minor; easy to read.
- `_build_agent_subprocess_env` logs a warning with `msg=...` as a structlog kwarg — consistent with the rest of the codebase, but worth confirming the Discord alert routing picks up warnings (no heartbeat on `agent-run` so this is log-only).
- The stale eventbridge.tf comment replacement is good; consider also removing the now-dead `fixed_scheduler_rules_map` reference from the `scheduled_jobs` output at line 278 if the map will stay empty for the foreseeable future. (Currently it joins an empty map, which is harmless but noisy.)

### Verification checklist from the issue
- [x] Agent `--db` default flipped to `odds`, docstring updated
- [x] `score_events` inlined + wrapped, outside `job_alert_context`
- [x] `_JOB_CRON_MAP` + `schedule_cron` added, CLI bootstrap branches on it
- [x] `bootstrap_jobs` default promoted, `test_config.py` updated
- [x] Terraform `fixed_schedule_expressions` emptied; `fetch-odds` untouched
- [x] `.env.example` adds `MODEL_NAME`/`MODEL_BUCKET`/`AGENT_DATABASE_URL`
- [x] `odds_agent` role migration (grants + reversible downgrade)
- [~] No new tests for cron bootstrap / allowlist gating — issue lists this implicitly under "success criteria" but the PR doesn't include them.

Non-blocking overall. Concerns 1 and 3 are the ones I'd want addressed before this lands.

PR #358, opened 2026-04-21
Move odds_agent role out of alembic; promote MODEL_NAME/BUCKET to Settings

Acusick1:
## Summary

- **Delete the `odds_agent` migration.** Roles and grants aren't schema — they're environment-level, and bundling them in alembic forced awkward ordering constraints (the APScheduler grants depend on tables that are only created when the scheduler boots, not by alembic). Provision the role manually per environment using the SQL snippet that's now in `.env.example`. The previous migration was not yet applied anywhere (local head still at `beaf16386050`, no successful deploy since PR #357 merged), so deleting the file is safe — nothing to reconcile.
- **Promote `MODEL_NAME` / `MODEL_BUCKET` to `Settings`.** Adds a `ModelConfig` pydantic block; `model_loader`, `score_predictions`, and `daily_digest` now read `settings.model.name` / `settings.model.bucket` instead of calling `os.environ.get` directly. Brings these env vars under the same validated, cached, dotenv-loaded path as the rest of `Settings`.
- **Surface `get_team_context` in the EPL agent prompt.** The tool landed in #352 but wasn't referenced anywhere in `common.md` or `epl/base.md`, and the most recent EPL run never reached for it. Added a short bullet under "Sport-Mechanical Workflow Notes" situating it next to `get_predictions` (fundamental-value reads around rotation, form, standings).

## Notes

- `.env.example` comment block trimmed to the bare operator-facing essentials (what the env var does, the SQL to provision the role, ordering constraint re. APScheduler tables). Previous version was too verbose and referenced the migration history.
- Pre-existing `test_config.py::test_settings_defaults` failure (Discord webhook leak from the local `.env`) still present; unrelated to this change.


Acusick1 (commented):
Reviewed the diff against `~/.claude/review-checklist.md`. CI is green and the three moves are independent and well-scoped.

**Verified**
- Migration deletion is safe: no in-tree references to `f1a2b3c4d5e6` remain, and `beaf16386050` is still the alembic head (`migrations/versions/beaf16386050_add_state_column_to_espn_fixtures.py`). Nothing to reconcile.
- Terraform (`deployment/aws/terraform/lambda.tf:119-120`) still injects `MODEL_BUCKET` / `MODEL_NAME` into the Lambda env, and `ModelConfig(env_prefix="MODEL_")` reads them transparently — no deploy-side regression.
- `AGENT_DATABASE_URL` agent-subprocess flow in `odds_lambda/jobs/agent_run.py` is unchanged; the role provisioning SQL now lives in `.env.example` as described.
- `epl/base.md` bullet reads cleanly and slots under "Sport-Mechanical Workflow Notes" next to `get_predictions`, consistent with the tool's surface.

**Nits (non-blocking)**
1. Import placement inconsistency: `model_loader.py` imports `get_settings` at module top; `score_predictions.py:159` and `daily_digest.py:307` defer it inside the function body. Would hoist to top-level in both for consistency — there's no cycle to avoid.
2. `score_events()` resolves `model_name` via `settings.model.name` but leaves `bucket` to fall through to `load_model()`. Two layers of the same fallback — could hoist both or neither; minor.
3. `tests/unit/test_model_loader.py::_isolate_model_settings` uses `monkeypatch.setenv("MODEL_NAME", "")` to neutralise a local `.env`. `monkeypatch.delenv(..., raising=False)` would be semantically cleaner (unset vs. empty string). Current form works because `not ""` is truthy in `load_model`'s guard.
4. `ModelConfig` sets `protected_namespaces=()` (`config.py:122`). Its fields (`name`, `bucket`) don't start with `model_`, so the override is only actually required on the parent `Settings` class (which gains a field literally named `model`). Harmless but slightly redundant on the child.

Pre-existing `test_config.py::test_settings_defaults` Discord-webhook flake is noted in the PR body — confirmed unrelated.

LGTM modulo the nits.

PR #359, opened 2026-04-22
Use XGBoostAdapter in score_predictions for multi-feature-group scoring

### LINKED ISSUE #354: Upgrade score_predictions feature extraction to use XGBoostAdapter for multi-feature-group scoring
## Goal

The production scorer in `packages/odds-lambda/odds_lambda/jobs/score_predictions.py` hand-rolls a tabular-only feature vector and will throw `feature_name_mismatch` for any model that uses additional feature groups. Replacing the extraction with the same `collect_event_data` + `XGBoostAdapter` machinery already used by training unlocks shipping `standings`, `match_stats`, and `epl_schedule` models to production with zero risk to the currently-deployed tabular-only `epl-clv-home`.

## Scope

- Replace `_extract_features` (`score_predictions.py:80-131`) with a two-call path: `collect_event_data(...)` to build an `EventDataBundle`, then `XGBoostAdapter().transform(bundle, snapshot, config)`.
- Preload the caches required by `config.feature_groups` once per `score_events()` invocation and pass them into `collect_event_data`. Mirror the guarded-preload pattern from `odds_analytics/feature_groups.py:1117-1141`.
- Validate feature-name alignment by comparing `XGBoostAdapter().feature_names(config)` against the model artifact's bundled `feature_names`; log `feature_name_mismatch` and skip when they diverge (preserves today's guard).
- Drop the `np.nan_to_num(..., nan=0.0)` step so the scorer matches training (XGBoost handles NaN natively; `nan_to_num` was removed from training in #231).

## Location

- Primary: `packages/odds-lambda/odds_lambda/jobs/score_predictions.py`
- Reference pattern: `packages/odds-analytics/odds_analytics/feature_groups.py:1117-1153` (preload + per-event `collect_event_data`)
- Adapter: `odds_analytics.feature_groups.XGBoostAdapter` (`feature_groups.py:791-810`)

## Implementation approach

1. In `score_events`, after resolving `config` and `sport_key` but before the event loop, conditionally load `standings_cache`, `match_stats_cache`, `fixtures_df` (and `lineup_cache` if/when relevant) based on `config.feature_groups` — gated exactly as training does it.
2. Inside the event loop, call `bundle = await collect_event_data(event, session, config, standings_cache=..., match_stats_cache=..., fixtures_df=..., lineup_cache=...)` once per event, then `adapter.transform(bundle, snapshot, config)` per unscored snapshot.
3. Skip the snapshot when `adapter.transform(...)` returns `None` (existing skip semantics: missing odds, etc.).
4. Use `AdapterOutput.features` as the `np.float32` feature vector passed to `model.predict(...)`.

## Success criteria

- [ ] `_extract_features` helper is removed (or reduced to a thin wrapper around the adapter call).
- [ ] `score_events` loads caches once per invocation, gated on `config.feature_groups`.
- [ ] Scoring with `feature_groups=['tabular']` produces byte-identical feature vectors to the current extraction (modulo the dropped `nan_to_num`, which is a no-op on tabular in practice).
- [ ] Scoring with `feature_groups=['tabular', 'standings']` succeeds end-to-end in a local smoke test against a dev-DB event and writes a `Prediction` row.
- [ ] Feature-name-mismatch guard still logs and skips (not raises) when the adapter's names diverge from the model's bundled names.
- [ ] Unit tests cover: (a) tabular-only path still produces the expected vector, (b) multi-group path succeeds when caches are provided, (c) name-mismatch path returns `None` / skips cleanly.

## Notes

- `epl_lineup` and `polymarket` feature groups have other data-availability blockers in prod and are out of scope for verification here — but the adapter path should handle them transparently if a future model requests them.
- `_load_fixtures_df` and `_load_lineup_cache` in `feature_groups.py` are module-private; either import with the leading underscore or promote them — implementer's call.
- Zero-risk deploy: the currently-deployed `epl-clv-home` (tabular-only) produces the same feature vector under the new path.

Acusick1:
## Summary
- Replace hand-rolled `_extract_features` in `score_predictions.py` with `collect_event_data` + `XGBoostAdapter().transform(...)`, mirroring training. Unblocks shipping `standings`, `match_stats`, `epl_schedule` models with zero risk to the deployed tabular-only `epl-clv-home`.
- Conditionally preload `standings_cache`, `match_stats_cache`, `fixtures_df`, `lineup_cache` once per `score_events()` invocation, gated on `config.feature_groups` — mirrors training preload pattern at `feature_groups.py:1117-1141`.
- Move the `feature_name_mismatch` guard from per-snapshot to once-per-invocation (adapter feature names depend only on `config`, not snapshots — preserves log-and-skip semantics, removes per-snapshot log noise).
- Drop `np.nan_to_num(..., nan=0.0)` to match training (XGBoost handles NaN natively after #231).
- Loosen `BacktestEvent.home_score` / `away_score` to `int | None` so the adapter can run on `SCHEDULED` events at inference time. All existing consumers in `backtesting/services.py` are reached only after `BacktestEvent.from_db_event` filters out scoreless events.

## Test plan
- [x] Unit tests: tabular-only adapter output is byte-identical to the legacy hand-rolled vector (modulo dropped `nan_to_num`).
- [x] Unit tests: multi-group path (tabular + standings) succeeds end-to-end.
- [x] Unit tests: `feature_name_mismatch` path returns empty stats and skips cleanly.
- [x] Unit tests: `_preload_caches` gating verified across 6 cases (tabular only, standings, match_stats, epl_schedule, epl_lineup, all groups).
- [x] CI green (1878 unit tests + integration tests pass).
- [x] **Live smoke test against dev DB**: ran `score_events` with `feature_groups=("tabular", "standings")` against a stub model (XGBoostAdapter feature_names asserted as 18: 6 tabular + 11 standings + 1 hours_until_event). Result: 19 upcoming EPL events checked, 1134 snapshots scored, 905 skipped (missing odds — expected), 0 errors. Standings cache loaded (11 seasons, 4103 events). Prediction rows written and cleaned up afterwards.

## Closes #354

### DIFF
--- a/packages/odds-analytics/odds_analytics/backtesting/models.py
+++ b/packages/odds-analytics/odds_analytics/backtesting/models.py
@@ -25,14 +25,21 @@ __all__ = [
 
 
 class BacktestEvent(BaseModel):
-    """Event validated for backtesting - guaranteed to have final scores."""
+    """Event payload used for both backtesting and live scoring.
+
+    Constructed via ``from_db_event`` for completed events (scores guaranteed
+    non-None) and via ``feature_groups.make_backtest_event`` for upcoming
+    events at inference time (scores may be ``None``). Score fields are
+    therefore optional; backtest call sites that read them are reached only
+    after ``from_db_event`` has filtered out scoreless events.
+    """
 
     id: str
     commence_time: datetime
     home_team: str
     away_team: str
-    home_score: int
-    away_score: int
+    home_score: int | None
+    away_score: int | None
     status: EventStatus
 
     @field_validator("commence_time")
--- a/packages/odds-analytics/odds_analytics/backtesting/services.py
+++ b/packages/odds-analytics/odds_analytics/backtesting/services.py
@@ -41,6 +41,20 @@ __all__ = ["BettingStrategy", "BacktestEngine"]
 logger = structlog.get_logger()
 
 
+def _require_scores(event: BacktestEvent) -> tuple[int, int]:
+    """Narrow ``BacktestEvent`` scores to non-None for outcome evaluation.
+
+    ``BacktestEvent.home_score`` / ``away_score`` are typed ``int | None`` so
+    the same model can carry pre-match events through feature extraction, but
+    all bet-evaluation paths must run on completed events. Raises
+    ``AssertionError`` if a caller reaches here with missing scores.
+    """
+    assert event.home_score is not None and event.away_score is not None, (
+        f"BacktestEvent {event.id} reached outcome evaluation without final scores"
+    )
+    return event.home_score, event.away_score
+
+
 class BettingStrategy(ABC):
     """Abstract base class for all betting strategies."""
 
@@ -304,7 +318,8 @@ class BacktestEngine:
         wins, backing a team loses. For 2-way markets (basketball), a tie is
         a push (returns None).
         """
-        side = determine_h2h_winner(event.home_score, event.away_score)
+        home_score, away_score = _require_scores(event)
+        side = determine_h2h_winner(home_score, away_score)
         if side == "draw":
             if is_three_way:
                 return outcome.lower() == "draw"
@@ -320,15 +335,16 @@ class BacktestEngine:
         if line is None:
             return None
 
+        home_score, away_score = _require_scores(event)
         if outcome == event.home_team:
-            adjusted_score = event.home_score + line
-            won = adjusted_score > event.away_score
-            push = adjusted_score == event.away_score
+            adjusted_score = home_score + line
+            won = adjusted_score > away_score
+            push = adjusted_score == away_score
         else:
             away_line = -line
-            adjusted_score = event.away_score + away_line
-            won = adjusted_score > event.home_score
-            push = adjusted_score == event.home_score
+            adjusted_score = away_score + away_line
+            won = adjusted_score > home_score
+            push = adjusted_score == home_score
 
         if push:
             return None
@@ -342,7 +358,8 @@ class BacktestEngine:
         if line is None:
             return None
 
-        total_points = event.home_score + event.away_score
+        home_score, away_score = _require_scores(event)
+        total_points = home_score + away_score
 
         if outcome.lower() == "over":
             won = total_points > line
--- a/packages/odds-analytics/odds_analytics/feature_groups.py
+++ b/packages/odds-analytics/odds_analytics/feature_groups.py
@@ -80,6 +80,10 @@ __all__ = [
     "filter_completed_events",
     "resolve_outcome_name",
     "snapshot_has_bookmaker",
+    "FeatureGroupCaches",
+    "preload_feature_group_caches",
+    "load_fixtures_df",
+    "load_lineup_cache",
 ]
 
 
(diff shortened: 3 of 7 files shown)


Acusick1 (commented):
## Review — PR #359

Solid, well-scoped implementation of #354. The scorer now shares the same `collect_event_data` → `XGBoostAdapter.transform` path as training, byte-identical for tabular-only and trivially extensible to richer feature groups. Test coverage is strong (byte-equality vs legacy, multi-group, name-mismatch, six preload-gating cases), CI is green, and the live smoke test in the PR description covers the actual code path.

### What checks out
- `_preload_caches` faithfully mirrors `prepare_training_data:1123-1146` — same guard set, same loaders. `sport_key` is validated as non-empty before `_preload_caches` is called (`score_predictions.py:175-181`), so dropping the defensive `if sport_key:` guard is safe.
- `feature_name_mismatch` hoisted out of the snapshot loop is correct — `XGBoostAdapter.feature_names(config)` depends only on `config`, not on any `bundle` or `snapshot`.
- `BacktestEvent` scores → `int | None` is semantically safe. Every runtime read of `home_score`/`away_score` in `backtesting/services.py` (e.g. `_evaluate_moneyline_outcome`, `_evaluate_spread_outcome`, `_evaluate_total_outcome`) is reached only after `from_db_event` filters scoreless events, and no other construction site passes `None` at inference time except the new `make_backtest_event`. Feature extractors don't read scores.
- Dropping `nan_to_num` aligns with the post-#231 training convention.

### Suggestions (non-blocking)

1. **Private-API import crosses a boundary.** `_load_fixtures_df` and `_load_lineup_cache` are module-private in `odds_analytics.feature_groups` but now have two callers across two packages. The issue explicitly flagged this — I'd promote both to public (`load_fixtures_df`, `load_lineup_cache`) in a follow-up since the underscore convention no longer reflects reality.

2. **Preload logic is duplicated.** `_preload_caches` (here) and the inline preload in `prepare_training_data:1122-1146` encode the same gating rules. Adding a new feature group with a cache now requires updating both. A shared helper in `feature_groups.py` returning the 4-tuple (or a small dataclass) would keep them in lockstep. Worth a follow-up rather than a blocker.

3. **Typing debt from `int | None` scores.** `services.py:324-346` does `event.home_score + line`, `event.home_score + event.away_score`, etc. without None-guards. Runtime-safe today, but a strict type-checker would flag it, and a future code path that constructs `BacktestEvent` at inference time and then reaches `_evaluate_*` would silently raise `TypeError`. Either tighten the type back to `int` on the backtest hot path (e.g. a separate `ScoredBacktestEvent` subtype) or add explicit asserts in `_evaluate_*`. Low priority.

4. **Minor — return shape.** `_preload_caches` returns a 4-tuple that callers unpack positionally. If this grows to 5+ entries a `NamedTuple`/dataclass will age better; 4 is still fine.

### Nits
- `_preload_caches` docstring references `prepare_training_data`'s "guarded-preload pattern" — that's a good orienting pointer for the next reader; keep it.

No blockers from my side — once the private-import question is acknowledged (or a follow-up issue filed), this is good to merge.

_(Comment-only: PR is co-authored by Claude, so I cannot approve.)_

PR #363, opened 2026-04-25
Add direct Betfair Exchange ingestion for EPL and MLB

Acusick1:
## Summary

- Restores `betfair_exchange` as a sharp benchmark by ingesting Match Odds directly via the Betfair Exchange API for both EPL (1x2) and MLB (h2h), replacing the indirect-via-OddsPortal path that broke 2026-04-21 when OP moved exchange odds to a separate encrypted endpoint.
- New `odds_lambda.betfair` package (cert-based auth, sport-parameterised event/market-book discovery, adapter into the canonical `bookmakers[].key=\"betfair_exchange\"` shape) plus a self-scheduling `fetch_betfair_exchange` job wired into the per-sport scheduler for `soccer_epl` and `baseball_mlb`.
- Existing consumers (`find_retail_edges`, hybrid Pinnacle+BFE sharp fallback) pick up the new rows unchanged via the canonical bookmaker key.

## Architecture notes

- **Auth**: cert-based non-interactive login (`identitysso-cert.betfair.com`) when `BETFAIR_CERT_FILE`/`BETFAIR_CERT_KEY` are set; falls back to `login_interactive` otherwise (dev only — prompts 2FA). `.gitignore` excludes `.betfair/` and `secrets/` so the cert pair can never be committed.
- **Cadence**: `*/30 * * * *` safety-floor cron, with proximity-based self-scheduling (15min closing tier `<3h` to KO, 30min pregame tier 3–12h, 2h far tier). Pre-schedules **before** fetching so a crash mid-run can't break the chain.
- **Storage**: writes a single `bookmakers[]` entry with `key=\"betfair_exchange\"`, `markets[0].key=\"1x2\"` for EPL or `\"h2h\"` for MLB, American prices via `decimal_to_american`, plus a `betfair_meta` block (`market_id`, `total_matched`, per-runner LTP+sizes). `raw_data.source=\"betfair_api\"` for provenance. Uses `find_or_create_event(home, away, commence_time)` against existing `op_live_*` events.
- **Bootstrap default updated**: `fetch-betfair-exchange` added to `SchedulerConfig.bootstrap_jobs` so the local scheduler installs both per-sport cron entries on start.

## Verified end-to-end against local DB

- **Cert login**: works against the live API with the kwargs corrected (`cert_files=(crt, key)`; the `betfairlightweight` `certs=` kwarg is a directory, not a tuple).
- **EPL**: 6/6 upcoming events ingested, 3-way prices (e.g. `Man Utd −112 / Brentford +330 / Draw +300`, `total_matched=£24,728`).
- **MLB**: 13/13 today's games ingested, 2-way h2h with home/away swapped to canonical order. Pitcher `(TBD)` annotations stripped.
- **In-play / closed** markets correctly skipped. **SUSPENDED** markets are kept (BFE state during brief trading pauses; price still valid pre-match).
- Self-scheduling tightened to 15min for MLB (games <2h away) and 2h for EPL (next fixture Mon Apr 27 19:00).

## Test plan

- [x] Unit tests: 43/44 pass (the single failure is the pre-existing `test_settings_defaults` Discord webhook leak documented in memory; my bootstrap assertion update passes).
- [x] Live cert login + listEvents/listMarketBook against the Betfair API (EPL+MLB).
- [x] Job `main()` invoked end-to-end against local Postgres for both sports; `odds_snapshots` rows verified for shape, prices, `total_matched`, and `betfair_meta`.
- [x] `ruff check` + `ruff format` clean.
- [ ] After merge: set `BETFAIR_CERT_FILE`/`BETFAIR_CERT_KEY` to **absolute** paths in the prod env (Lambda cwd is `/var/task`, the local dev `.betfair/...` relative path won't resolve).
- [ ] After merge: confirm `betfair_login_cert` (not `betfair_login_interactive`) appears in CloudWatch on the first prod run.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,8 +15,8 @@ SCHEDULER_LOOKAHEAD_DAYS=7
 # the fixed-schedule jobs (daily-digest, fetch-espn-fixtures) that now run
 # on the local scheduler instead of EventBridge.
 # Default: ["agent-run", "fetch-oddsportal", "fetch-oddsportal-results",
-#           "daily-digest", "fetch-espn-fixtures"]
-# SCHEDULER_BOOTSTRAP_JOBS=fetch-oddsportal,fetch-oddsportal-results,daily-digest,fetch-espn-fixtures,agent-run
+#           "daily-digest", "fetch-espn-fixtures", "fetch-betfair-exchange"]
+# SCHEDULER_BOOTSTRAP_JOBS=fetch-oddsportal,fetch-oddsportal-results,daily-digest,fetch-espn-fixtures,agent-run,fetch-betfair-exchange
 
 # =============================================================================
 # AWS CONFIGURATION (only needed if SCHEDULER_BACKEND=aws)
@@ -145,6 +145,33 @@ DATA_QUALITY_ERROR_THRESHOLD=10
 # Alert rate limiting
 # Minimum minutes between identical alerts (default: 30)
 ALERT_RATE_LIMIT_MINUTES=30
+
+# =============================================================================
+# BETFAIR EXCHANGE API (read-only sharp benchmark)
+# =============================================================================
+# Direct ingestion of Match Odds via the free delayed application key.
+# Replaces the indirect-via-OddsPortal BFE path. Sport scope: EPL + MLB.
+#
+# Cert-based non-interactive login is the production path (bypasses 2FA).
+# Leaving cert paths unset falls back to interactive login (dev-only;
+# prompts 2FA if enabled on the account).
+
+# Toggle ingestion. Defaults to false; set to true once credentials are in place.
+BETFAIR_ENABLED=false
+
+# Account credentials and delayed app key
+BETFAIR_USERNAME=
+BETFAIR_PASSWORD=
+BETFAIR_APP_KEY=
+
+# Optional: cert paths for non-interactive login. Generate via XCA / openssl
+# and upload the .crt to your Betfair account at developer.betfair.com.
+# Cert lifetime is 365 days; rotate annually.
+BETFAIR_CERT_FILE=
+BETFAIR_CERT_KEY=
+
+# Look-ahead window for event discovery (hours)
+BETFAIR_LOOKAHEAD_HOURS=168
 
 # =============================================================================
 # LOGGING
--- a/.env.example
+++ b/.env.example
@@ -173,6 +173,10 @@ BETFAIR_CERT_KEY=
 # Look-ahead window for event discovery (hours)
 BETFAIR_LOOKAHEAD_HOURS=168
 
+# Optional: comma-separated sport scope override
+# (default: every sport supported by the BFE adapter; currently EPL + MLB)
+# BETFAIR_SPORTS=soccer_epl,baseball_mlb
+
 # =============================================================================
 # LOGGING
 # =============================================================================
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,10 @@ ENV/
 .env.local
 .env.prod
 
+# Betfair Exchange API certificates (never commit private keys)
+.betfair/
+secrets/
+
 # IDE
 .vscode/
 .idea/
--- a/packages/odds-core/odds_core/config.py
+++ b/packages/odds-core/odds_core/config.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from functools import lru_cache
 from pathlib import Path
 
-from pydantic import Field
+from pydantic import Field, model_validator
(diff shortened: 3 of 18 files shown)


Acusick1 (commented):
## Review

Direct BFE ingestion looks well-architected and follows the existing fetch_oddsportal patterns closely. Verification artefacts in the description (live cert login, 6/6 EPL + 13/13 MLB ingest, total_matched + meta inspected) cover the integration risk that's hard to exercise in unit tests. CI green; 43/44 unit tests pass with the documented pre-existing failure.

### Architecture / data shape
- `key="betfair_exchange"` + sport-specific `market_key` (`1x2` / `h2h`) + canonical team names means downstream consumers (`find_retail_edges`, hybrid Pinnacle+BFE sharp fallback) pick this up unchanged. No migration or schema changes required.
- Cert-based login with the `cert_files=(crt, key)` tuple kwarg is correct (the `betfairlightweight` `certs=` kwarg takes a directory). `.gitignore` excludes `.betfair/` and `secrets/` — good guardrail.
- Pre-schedule before fetch + `*/30` safety floor matches the `fetch_oddsportal` survivability story.
- Free delayed app key = no quota cost, so the `*/30` floor running on idle days is acceptable.

### Observations / minor issues

1. **Pre-schedule failure handling diverges from `fetch_oddsportal`.** `fetch_oddsportal` raises on pre-schedule failure (`packages/odds-lambda/odds_lambda/jobs/fetch_oddsportal.py:300`); the new job logs and continues (`fetch_betfair_exchange.py:1093-1094`). The `*/30` floor means a failed self-schedule still gets caught within 30min, so this is probably fine — but worth deciding consciously rather than diverging by accident.

2. **`enabled=True` + missing creds is a soft failure.** `BetfairConfig` doesn't enforce that creds are set when `enabled=True`; the job logs `betfair_credentials_missing` and returns silently. A Pydantic `model_validator` would surface a misconfigured prod env at startup rather than at the next cron tick. Minor.

3. **EPL `competition_id` has no name-hint fallback.** `MLB_COMPETITION_NAME_HINTS` provides a discovery fallback if Betfair re-issues the ID; EPL relies on the pinned `10932509` only. If Betfair ever re-issues it, the EPL job goes silently empty (the soft warning catches it eventually). Adding `competition_name_hints=("Premier League",)` to the EPL config in `constants.py:111-119` would mirror the MLB resilience for free.

4. **`SUSPENDED` markets pass through.** Per the PR description this is intentional (transient pause, prices still valid pre-match). Consider counting them in `IngestionStats` so they show up in the sport-level summary log — the current `books_in_play` counter already proves the pattern is useful.

5. **Adapter name-fallback is silent.** `_runner_to_outcome` (`adapter.py:282-290`) falls back to the canonical name when it doesn't match home/away — a defensive guard against alias-table gaps. But there's no log/counter, so a misaliased team would write a snapshot with three "outcomes" that don't all map to {home, away, Draw}. A `logger.warning` (or a stats counter) here would catch alias drift early. Low priority.

6. **`api_request_id` repeated across all snapshots in a run** (`fetch_betfair_exchange.py:1003`). This matches `fetch_oddsportal`'s pattern (`api_request_id = f"oddsportal_live_{scraped_date}"` shared per-run) so it's consistent.

7. **`BETFAIR_SPORTS` env is silently supported but undocumented in `.env.example`.** Pydantic's `env_prefix="BETFAIR_"` will pick up `BETFAIR_SPORTS=...` overrides, but the .env.example doesn't mention it. Trivial to add; not a blocker.

8. **`if TYPE_CHECKING: pass`** in `client.py:27-28` — looks like a vestigial block; safe to delete.

9. **Sport list duplication.** `BetfairConfig.sports` (config), `SPORT_CONFIG` keys (constants), and the `SCHEDULE_OVERRIDES` tuple in `scheduling/jobs.py:1188` all enumerate the same sports. Adding a third sport requires touching all three. Not new tech debt — same shape as `_LEAGUE_SPEC_BY_SPORT` in `fetch_oddsportal` — but worth flagging if more sports land.

### Type/style
- All new code uses modern type hints (`str | None`, `list[T]`), `datetime.now(UTC)`, async-context-managed sessions. Clean.
- Frozen dataclasses for the wire-shape DTOs (`BetfairEvent`, `BetfairBook`, `BetfairRunner`) is the right call.

### Testing
- Adapter coverage (alias/normalization/draw/pitcher-strip/parse-fail/missing-outcome) is solid.
- `_should_skip_book` and `_interval_for_kickoff` covered. No test for the `find_or_create_event` 2h-window join with existing OP-created events; verified live per the PR description, which is acceptable for an external-API shim.

Cannot approve — co-authored by Claude. Posting as a comment.

PR #366, opened 2026-04-25
Resolve retail/sharp prices via per-bookmaker lookback (#364)

### LINKED ISSUE #364: Resolve retail/sharp prices via per-bookmaker lookback so BFE-only snapshots don't hide retail data
## Goal

`find_retail_edges` and `get_current_odds` currently call `get_latest_snapshot()` and read all bookmaker prices from that single row. Pre-2026-04-21 a single OddsPortal scrape produced one snapshot containing both retail and sharp books, so this worked. Now that Betfair Exchange is ingested directly via `fetch_betfair_exchange` into a separate snapshot row, BFE often wins the \"latest snapshot\" race and retail data from the older OP snapshot becomes invisible to consumers. Switch both consumers to a per-bookmaker latest-within-lookback resolution so they're robust to multiple writer streams.

## Scope

- Add `OddsReader.get_latest_book_prices(event_id, market, lookback_hours, *, now=None)` returning, for each `bookmaker_key`, its most recent price (per outcome and point) within `[now − lookback_hours, now]`.
- Update `find_retail_edges` to source retail rows from the new reader instead of `get_latest_snapshot()`. Sharp resolution via `get_sharp_prices` is unchanged.
- Update `get_current_odds` to source bookmaker rows from the new reader instead of `get_latest_snapshot()`.
- Add a `lookback_hours: float = 2.0` parameter to both MCP tools so it's tunable per call.
- Surface per-row staleness:
  - `find_retail_edges` retail rows gain `snapshot_time` + `book_age_seconds` (mirroring the existing `sharp_snapshot_time` / `sharp_age_seconds`).
  - `get_current_odds` bookmaker entries already carry `last_update` from `raw_data`; additionally surface the contributing snapshot's `snapshot_time` per bookmaker.
- Top-level `snapshot_time` on both responses becomes \"newest contributing snapshot in the window\" (max of surfaced book snapshot times). Document the new semantics in the docstrings.

## Success criteria

- [ ] `OddsReader.get_latest_book_prices` returns one entry per `(bookmaker_key, outcome_name, point)` taken from the newest snapshot in the window that contains that book; books absent from the window do not appear.
- [ ] When the latest snapshot for an event is a BFE-only row and an older OP snapshot in the window contains 17 retail books, `find_retail_edges` returns 17 retail books (filter still drops `pinnacle` / `betfair_exchange` from the retail set as today).
- [ ] When the same retail bookmaker appears in two snapshots within the window, only its newest price is surfaced.
- [ ] Each retail row in `find_retail_edges` carries `snapshot_time` and `book_age_seconds`; the top-level `snapshot_time` equals the max across all surfaced retail snapshot times.
- [ ] `get_current_odds` returns the most recent price per bookmaker across the window (not just the latest single snapshot), with per-bookmaker `snapshot_time` accessible to the agent.
- [ ] `lookback_hours` parameter on both tools defaults to `2.0` and accepts a float override.
- [ ] Existing dispersion stats (`median_divergence`, `dispersion_stddev`, `z_score`, `market_hold`) on `find_retail_edges` continue to work, computed across the per-book-latest set.
- [ ] Unit tests: (a) BFE row newer than OP row → 17 retail surfaced, (b) book A only in N−2 + book B only in N → both surfaced with correct `snapshot_time`, (c) book appears in both N−2 and N → only N's price surfaced, (d) book absent from window → not surfaced.
- [ ] Tool docstrings updated to describe the new lookback semantics and per-row staleness fields.

## Location

- `packages/odds-lambda/odds_lambda/storage/readers.py` — new `get_latest_book_prices` (mirror the structure of `get_sharp_prices` at line 637).
- `packages/odds-mcp/odds_mcp/server.py` — `find_retail_edges` (line 767) and `get_current_odds` (line 191).
- `tests/unit/` — new tests for the reader; update existing `find_retail_edges` / `get_current_odds` tests if any reference the old single-snapshot semantics.

## Implementation approach

Mirror `get_sharp_prices`: query all `OddsSnapshot` rows for the event in `[now − lookback_hours, now]` newest-first, walk them, and for each `(bookmaker_key, outcome_name, point)` keep the first (newest) price seen. Reuse `extract_odds_from_snapshot` and `raw_data_has_market` for consistency.

Keep `get_latest_book_prices` separate from `get_sharp_prices`. They share the lookback query shape but their dedup semantics differ (sharp does priority-ordered per-outcome fallback and can stop early; the new reader keeps one row per `(bookmaker_key, outcome, point)` and must walk the whole window). A shared private helper for the window query is fine if it falls out naturally; not required.

Pre-compute `age_seconds` inside the reader, anchored on a `now: datetime | None = None` kwarg defaulting to `datetime.now(UTC)`. Return a structure carrying `snapshot_id`, `snapshot_time`, `age_seconds` per book — mirrors `SharpPriceMeta` so consumers handle staleness uniformly across sharp and retail. `find_retail_edges` passes `now=<latest_snapshot_time>` for the data-relative window anchor (same pattern it uses for sharp today).

`find_retail_edges` keeps its current sharp-set filter (`if o.bookmaker_key in sharp_set: continue`) — that logic is unchanged and correct. Only the source of `odds` changes: from `extract_odds_from_snapshot(latest_snapshot)` to a flat list built from the new reader.

## Constraints

- Async DB access via `async_session_maker()`; no sync sessions.
- No new dependencies.
- Dispersion stats are now computed across non-co-temporal prices. This is acceptable (small distortion vs. line moves within a 2h pre-match window) but document the trade-off in the `find_retail_edges` docstring so the agent can weight `dispersion_stddev` appropriately.
- Do not pre-emptively unify with `get_sharp_prices`. The dedup semantics differ enough that a unified reader would be more complex than two focused ones.

## Notes

- The bug only surfaces post-2026-04-25 once `fetch_betfair_exchange` starts writing to local/prod; pre-existing snapshots have no BFE-only rows and `find_retail_edges` was working correctly for them.
- `_DEFAULT_SHARP_BOOKMAKERS` (`server.py:57`) is unchanged; this issue does not touch sharp resolution.
- Related: PR #363 introduced the BFE direct-ingestion path that surfaces this issue.

Acusick1:
## Summary

- New `OddsReader.get_latest_book_prices(event_id, market, lookback_hours, *, now=None)` resolves prices per `(bookmaker_key, outcome_name, point)` from the newest snapshot containing each book within `[now − lookback_hours, now]`. Mirrors the structure of `get_sharp_prices` and reuses `extract_odds_from_snapshot` / `raw_data_has_market`. New dataclasses `BookPriceMeta`, `BookPriceEntry`, `BookPriceResult` carry per-book provenance (`snapshot_id`, `snapshot_time`, `age_seconds`).
- `find_retail_edges` and `get_current_odds` rewired to source rows from the new reader instead of `get_latest_snapshot()`. Each retail/bookmaker row now carries `snapshot_time` + `book_age_seconds` (mirroring existing sharp staleness fields), and the top-level `snapshot_time` is the max across surfaced book snapshot times. Both tools take a tunable `lookback_hours: float = 2.0` parameter.
- `get_current_odds` envelope: anchor-derived fields renamed to `anchor_snapshot_id`, `anchor_created_at`, `anchor_fetch_tier`, `anchor_hours_until_commence` to make scope explicit. `bookmaker_count` now derived from surfaced rows. `_odds_to_dict` exposes per-row `last_update` from `raw_data`.
- Sharp resolution via `get_sharp_prices` is unchanged; `find_retail_edges` keeps its sharp-set filter.
- Docstrings updated to describe new lookback semantics, per-row staleness, and the trade-off that dispersion stats are now computed across non-co-temporal prices (acceptable within a 2h pre-match window).

## Why

Before PR #363, OddsPortal scrapes wrote single snapshots containing both retail and sharp books, so reading the latest snapshot row surfaced everything. With direct Betfair Exchange ingestion (`fetch_betfair_exchange`), BFE-only snapshots can win the latest-snapshot race and silently hide the older OP snapshot's retail data from `find_retail_edges` and `get_current_odds`. Per-bookmaker lookback resolution makes both consumers robust to multiple writer streams.

## Tests

- New unit tests in `tests/unit/test_get_latest_book_prices.py` cover the four spec scenarios (BFE-newer surfaces 17 retail; books only in older/newer both surface; same book in both → newest only; out-of-window not surfaced) plus market filtering, age anchoring, and totals point keying.
- Integration tests in `tests/integration/test_mcp_tools.py` exercise BFE-newer end-to-end for both tools, plus `lookback_hours` override, totals multi-point, `last_update` propagation, anchor field semantics, `bookmaker_count` derivation, and `include_raw_data`.

## Closes #364

### DIFF
--- a/packages/odds-lambda/odds_lambda/storage/readers.py
+++ b/packages/odds-lambda/odds_lambda/storage/readers.py
@@ -828,10 +828,17 @@ class OddsReader:
 
             age_seconds = round((ref_time - snapshot.snapshot_time).total_seconds(), 1)
 
+            # Dedup is first-wins both across and within snapshots: snapshots
+            # are walked newest-first, so the first time a
+            # ``(bookmaker_key, outcome_name, point)`` tuple is seen its price
+            # is locked in. Within-snapshot duplicates (the same tuple
+            # repeated inside one bookmaker block) are not expected in
+            # well-formed ``raw_data``; if they ever occur the earlier entry
+            # in the iteration order wins.
             for o in odds:
                 key = (o.bookmaker_key, o.outcome_name, o.point)
                 if key in book_result.entries:
-                    continue  # already resolved from a newer snapshot
+                    continue
                 book_result.entries[key] = BookPriceEntry(
                     odds=o,
                     meta=BookPriceMeta(
--- a/packages/odds-lambda/odds_lambda/storage/readers.py
+++ b/packages/odds-lambda/odds_lambda/storage/readers.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from dataclasses import dataclass, field
 from datetime import UTC, datetime, timedelta
 
 import structlog
@@ -16,6 +17,40 @@ from sqlalchemy.ext.asyncio import AsyncSession
 from odds_lambda.fetch_tier import FetchTier
 
 logger = structlog.get_logger()
+
+
+@dataclass
+class BookPriceMeta:
+    """Metadata about a per-bookmaker price found via lookback search.
+
+    Mirrors :class:`SharpPriceMeta` so consumers can handle staleness
+    uniformly across sharp and retail prices.
+    """
+
+    snapshot_id: int
+    snapshot_time: datetime
+    age_seconds: float
+
+
+@dataclass
+class BookPriceEntry:
+    """A single per-bookmaker price with its provenance."""
+
+    odds: Odds
+    meta: BookPriceMeta
+
+
+@dataclass
+class BookPriceResult:
+    """Per-bookmaker prices resolved across a time window of snapshots.
+
+    ``entries`` maps ``(bookmaker_key, outcome_name, point)`` to the newest
+    price for that tuple within the lookback window, along with provenance
+    metadata. Tuples for which no snapshot in the window contained the book
+    are absent from the mapping.
+    """
+
+    entries: dict[tuple[str, str, float | None], BookPriceEntry] = field(default_factory=dict)
 
 
 class OddsReader:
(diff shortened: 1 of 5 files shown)


Acusick1 (commented):
Reviewed against the checklist. Design looks correct: the anchor-on-`latest_snapshot.snapshot_time` pattern matches what `get_sharp_prices` already does, the dedup-on-`(bookmaker_key, outcome_name, point)` tuple is the right key for totals/spreads point splits, and the four spec scenarios from #364 are unit-covered. CI green.

A few non-blocking observations:

**1. `_make_book_result` mock comment is inverted (`tests/unit/test_mcp_server.py:680`).** The comment says *"Newer entries (later in the list) win — matches reader semantics"*, but the reader is **first-wins** under newest-first iteration (`if key in book_result.entries: continue` at `readers.py:838`). The mock is last-wins (always assigns). They're opposite. In practice unit tests pass odds with no duplicates so it doesn't bite, but the comment should either be corrected or the loop changed to `if key in entries: continue` for parity.

**2. Redundant `by_book` dedup in `find_retail_edges` (`server.py:1034-1036`).** `book_result.entries` is already keyed by `(bookmaker_key, outcome_name, point)`, so within a single `(outcome_name, point)` bucket every `bookmaker_key` is unique by construction. The `by_book: dict[str, BookPriceEntry]` rebuild is a no-op — `for entry in sorted(outcome_entries, key=lambda e: e.odds.bookmaker_key)` would be equivalent and clearer. Style only.

**3. Stale-price actionability risk under widened lookback isn't fully captured in the docstring.** The dispersion-stat caveat is good, but the bigger practical concern is that a retail row from a 60-min-old OP scrape (because BFE wrote a newer BFE-only snapshot in between) may no longer be takeable — the bookmaker can have moved its line. The agent does get `book_age_seconds` per row to judge this, but the default `lookback_hours=2.0` is generous relative to typical retail line moves. Worth one sentence nudging consumers to weight `book_age_seconds` when sizing, not just `dispersion_stddev`. Related: this overlaps with the strikethrough-as-live caveat in `project_scraper_strikethrough_gap.md` — both surface stale prices the agent has to discount.

**4. `get_current_odds` test for `lookback_hours` override is good; the equivalent for `find_retail_edges` is missing.** Minor — the reader-level test at `test_book_absent_from_window_not_surfaced` exercises the windowing, but a tool-level integration test confirming the override propagates through `find_retail_edges` would mirror the `get_current_odds` coverage.

**5. Breaking-change surface on `get_current_odds` envelope is real but contained.** Renames `snapshot_id` → `anchor_snapshot_id`, `created_at` → `anchor_created_at`, `fetch_tier` → `anchor_fetch_tier`, `hours_until_commence` → `anchor_hours_until_commence`, plus `bookmaker_count` semantics now derive from surfaced rows. No in-tree consumers other than the MCP tools themselves, so no code break — but the MLB agent observations log (`agents/mlb/observations.md:435-436`) records a heuristic where `bookmaker_count == 1` flagged "BFE-direct only" slates. Post-PR that heuristic still fires when the window contains only BFE, but if an older OP snapshot is in the 2h window the count will jump to 18+ — which is precisely the intended outcome of the fix, not a regression. Worth a quick prompt update so the agent isn't surprised by the new field names or by the better behavior. No PR change required, but flagging for the follow-up.

The four spec scenarios + market filter / age anchoring / totals-point keying give the reader strong coverage. Integration tests exercise the full BFE-newer-with-older-OP scenario through both tools end-to-end. LGTM modulo the comment fix in (1).

PR #374, opened 2026-05-29
Add live job-table printing and stale-schedule pruning to local scheduler

Acusick1:
## Summary
- Print the scheduled-job table at startup and on every `JobReleased` event, via a new `LocalSchedulerBackend.subscribe()` method that registers a listener on the running scheduler.
- Prune schedules that are no longer in `bootstrap_jobs` on start, so self-scheduling jobs removed from config stop firing after a restart (they otherwise persist indefinitely in the Postgres data store).
- Refactor the job-table rendering into a shared `_print_scheduled_jobs` helper reused by `scheduler list`.

## Testing
- `ruff check` / `ruff format` clean (pre-commit passed).

### DIFF
--- a/packages/odds-cli/odds_cli/commands/scheduler.py
+++ b/packages/odds-cli/odds_cli/commands/scheduler.py
@@ -1,6 +1,9 @@
 """Scheduler management CLI commands."""
 
+from __future__ import annotations
+
 import asyncio
+from typing import TYPE_CHECKING
 
 import structlog
 import typer
@@ -8,12 +11,15 @@ from odds_core.config import get_settings
 from rich.console import Console
 from rich.table import Table
 
+if TYPE_CHECKING:
+    from odds_lambda.scheduling.backends.base import ScheduledJob
+
 app = typer.Typer()
 console = Console()
 logger = structlog.get_logger()
 
 
-def _print_scheduled_jobs(con: Console, jobs: list) -> None:
+def _print_scheduled_jobs(con: Console, jobs: list[ScheduledJob]) -> None:
     if not jobs:
         con.print("[yellow]No jobs currently scheduled[/yellow]")
         return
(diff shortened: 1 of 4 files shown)


Acusick1 (commented):
## Review — local scheduler observability + stale-schedule pruning

Solid, well-scoped change. The live job table on `JobReleased` and the prune-on-restart are both genuinely useful for the local loop. Verified the mechanics against the apscheduler 4.0.0a6 source and the backend contract. Comment-only (Claude co-authored).

### CI
The single red test is `tests/unit/test_mlb_probable_pitchers_tool.py::test_write_through_returns_persisted_rows` (`assert 0 == 1` on a pglite write-through) — unrelated to this diff (touches no scheduler code) and looks like a DB-isolation flake. Please confirm it goes green on a rerun before merging; don't merge on a red board.

### Maintainability (main concern)
- **`_expected_compound_names()` duplicates the bootstrap loop's per-sport / cron / `target_sports` resolution** (scheduler.py:119-139 vs :221-263). These two now have to stay in lockstep — and the failure mode is asymmetric: if the bootstrap naming logic changes and the prune copy doesn't, the prune will compute a wrong `expected` set and **cancel live schedules**. Worth extracting a single `resolve_target_sports(job_name) -> list[str]` (or `expected_compound_names`) helper that both the prune and the bootstrap loop call, so they can't drift.
- **The prune logic is buried in a closure and has no test.** It's the one destructive operation added here (`cancel_scheduled_execution`), yet `_expected_compound_names` / the stale-filter can't be unit-tested as written. Extracting to module level (per above) would also make it testable — recommend a small test covering "sport removed from config → its compound is pruned" and "configured job is preserved."

### Minor
- `_print_scheduled_jobs(con: Console, jobs: list)` — bare `list`; should be `list[ScheduledJob]` (from `odds_lambda.scheduling.backends.base`) per the project's complete-type-hints rule.
- `_on_job_released` swallows all exceptions silently (`except Exception: return`). Display-only so low risk, but a `logger.debug(...)` would avoid a totally silent failure if `get_scheduled_jobs()` ever errors.
- The `JobReleased` listener reprints the full table on every job fire — can get noisy on a busy slate. Fine for a local dev tool, just flagging.

### Confirmed correct (no action)
- Async coroutine callback is supported — apscheduler ignores `is_async` for coroutine functions, so `_on_job_released` is fine; dropping the `Subscription` return is OK for the scheduler's lifetime.
- Sort key `(j.next_run_time is None, j.next_run_time)` is safe: two `None`s compare equal on the first element and never reach a `None < datetime` comparison.
- Prune is correctly gated on `effective_bootstrap`, so sidecar mode (`--db ... --no-bootstrap`) preserves manually-driven schedules. Good call.
- `subscribe()` wrapper with the `BackendUnavailableError` guard matches the backend's existing patterns.

PR #375, opened 2026-05-29
Lambda infra fixes: Betfair warning, log context, joblib dep

Acusick1:
## Summary
Three small, independent infrastructure fixes (one commit each):

- **Betfair empty-warning refinement** — only warn on zero stored snapshots when events were actually seen. An empty event list is a benign venue-side gap (e.g. next day's MLB markets not yet posted); warning only when events were returned but none stored targets genuine parsing/filtering regressions.
- **Lambda log context** — bind base job name and `aws_request_id` to the structlog context for all downstream log entries, and clear all contextvars on exit instead of unbinding only `sport`.
- **Declare `joblib`** — `model_loader` imports `joblib` directly; it was previously available only transitively.

## Testing
- `ruff check` / `ruff format` clean (pre-commit passed).


Acusick1 (commented):
## Review — PR #375 (Lambda infra fixes)

Three small, well-scoped, independent fixes (one commit each). All look correct. Comment-only because the commits are co-authored by Claude.

### 1. Betfair empty-warning refinement (`fetch_betfair_exchange.py`) ✅
Gating the soft warning on `total_stored == 0 and total_events_seen > 0` is the right call — it silences the benign "no markets posted yet" case while still flagging genuine parse/filter regressions (events returned but nothing stored). `events_seen` is populated on every `IngestionStats`, and the "no events / no books" early-return paths already log their own warnings, so coverage is intact.

**Minor (non-blocking):** when `ingest_sport` raises, the `except` builds `IngestionStats(sport=..., errors=[...])` with `events_seen=0`. So a *hard* per-sport ingestion failure now contributes `0` to `total_events_seen` and won't trip this aggregate warning — i.e. a run that errored out across all sports produces no "empty" warning here. That's defensible since the exception is logged at `error` level and the run is wrapped in `job_alert_context`, but worth a conscious confirmation that the per-sport error path is the intended owner of that signal.

### 2. Lambda log context (`lambda_handler.py`) ✅
Binding `job`/`request_id` and switching the `finally` from `unbind_contextvars("sport")` to `clear_contextvars()` is a real improvement: in a warm-reused Lambda container, clearing all contextvars prevents the newly-bound `job`/`request_id` (and `sport`) from leaking into a subsequent invocation. Good.

Two small notes:
- `context` is typed `object`, so `context.aws_request_id` is an untyped attribute access — fine at runtime (Lambda context always has it) but `ty`/static checkers may flag it. Consistent with the existing typing, so acceptable.
- Confirm `base_job_name` is in scope at the new binding site (it's referenced for sport resolution earlier in the handler, and CI would catch F821, so this is almost certainly fine — I couldn't re-verify locally due to a tooling glitch).

### 3. `joblib` explicit dependency (`pyproject.toml` + `uv.lock`) ✅
Correct — `model_loader` imports `joblib` directly, so it should be a declared dep rather than transitive. Lockfile updated consistently (both `dependencies` and `requires-dist`).

### Verdict
No correctness blockers. Clean separation of concerns across the three commits. (Note: I was unable to confirm CI status during review due to a local tool-output glitch — please confirm checks are green before merge.)

PR #380, opened 2026-05-31
Make get_probable_pitchers MCP tool read-only

### LINKED ISSUE #379: get_probable_pitchers: make MCP tool read-only; cron owns writes (fixes odds_agent permission failure)
## Problem

Every MLB agent run since 2026-05-09 hits a hard failure on `get_probable_pitchers`:

```
InsufficientPrivilegeError: permission denied for table mlb_probable_pitchers
[SQL: INSERT INTO mlb_probable_pitchers (...) ON CONFLICT ... DO NOTHING]
```

The agent subprocess connects as the read-mostly `odds_agent` role (via `AGENT_DATABASE_URL`, see `jobs/agent_run.py:139`). The `get_probable_pitchers` MCP tool is **write-through on every call** (`tools/mlb.py:94-98` → `MlbPitcherWriter.insert_snapshots` + `commit`), and `odds_agent` was never granted INSERT on `mlb_probable_pitchers`. So the tool errors instead of returning data, on every MLB slate. The agent works around it via RotoWire web research, but it's noisy, costs turns, and loses the cached-snapshot signal the tool is supposed to provide.

## Root cause is architectural, not a missing GRANT

`mlb_probable_pitchers` is **pipeline-ingested reference data** — conceptually identical to odds snapshots, fixtures, and predictions, all of which the agent *reads* but does not *write*. The agent role's legitimate writes are its own **outputs**: `paper_trade` and `match_brief`. Having a read-mostly agent role write-through to an ingestion table is the anomaly.

The cron's own docstring even declares the MCP tool the "primary write path" and the cron a "backstop" — that's backwards. Ingestion belongs to the scheduled job running under the write-capable role; the MCP tool should be a read surface.

(We should **not** just `GRANT INSERT ... TO odds_agent`. That patches the symptom and re-muddies the security boundary the read-mostly role exists to enforce — the agent would then be able to mutate reference-data tables.)

## Proposed change

1. **Make `get_probable_pitchers` read-only.** Drop the `MlbPitcherWriter` import, the `insert_snapshots` call, and the `commit` from `tools/mlb.py`. The agent role needs zero write grant on the table.
   - Keep live freshness via **read-through**: `refresh=True` live-fetches MLBAM and returns current state *without persisting*; on HTTP failure fall back to the cached read. `refresh=False` reads the cached snapshot only. External tool semantics are unchanged except that nothing is written. This preserves the agent's need for late-scratch-aware live data at T-2h while removing all writes from the agent path.
2. **`fetch-mlb-probables` cron becomes the sole writer** (it already runs under the write-capable role; #377). Update its docstring — it is the primary/only write path, not a backstop.

## Open decision (can defer)

The original rationale for write-through was a *dense announcement-timing series* (capturing the exact moment a pitcher is announced across an agent session). With the tool no longer writing, that series is owned solely by the cron at `0 6 * * *` — too coarse for dense timing. Options: (a) accept coarse daily series, since the agent reads live anyway and dense timing has no concrete consumer yet (recommended — don't over-build); or (b) bump cron cadence (e.g. every 2-3h during the announce window) if/when a research use materializes. Recommend (a) for now.

## Non-issue to confirm and close out

The agent also reported a "date-boundary bug" (schedule returning prior-day `game_pks` on game-day). This is **likely not real**: `dates_for_window` (`mlb_stats_fetcher.py:48`) over-fetches ±1 calendar day *by design* to absorb MLBAM's ET-anchored `date=` param, and `MlbPitcherReader.get_latest_in_window` filters returned rows by `commence_time ∈ [now, end]`. The `2026-05-30` rows the agent saw were in the **failed INSERT's leaked params** (full padded window), not in tool output. Worth a one-line confirmation, but no fix expected.

## Files

- `packages/odds-mcp/odds_mcp/tools/mlb.py` — remove write path
- `packages/odds-lambda/odds_lambda/jobs/fetch_mlb_probables.py` — docstring (sole writer)
- tests covering the tool's read/refresh behavior

Acusick1:
## Summary

Makes the MLB probable-pitcher data path read-only, then layers it cleanly. The MLB agent runs as the read-mostly `odds_agent` role with no INSERT on `mlb_probable_pitchers`, so the write-through `get_probable_pitchers` tool failed every MLB slate with `InsufficientPrivilegeError`. The fix is architectural: `mlb_probable_pitchers` is pipeline-ingested reference data the agent reads, not writes.

This PR carries three stacked layers (originally #379/#380 + #381 + a follow-on; the stack was collapsed into this single PR):

### 1. Read-only tool — closes #379

- **`packages/odds-mcp/odds_mcp/tools/mlb.py`** — tool is fully read-only. Dropped the writer, `insert_snapshots`, and `commit`. `refresh=True` read-throughs MLBAM live and returns current state per game without persisting; on `httpx.HTTPError` it falls back to the cached DB read (`stale_db_only`). `refresh=False` reads cached only (`db_only`).
- **`packages/odds-lambda/odds_lambda/jobs/fetch_mlb_probables.py`** — docstring declares the daily cron the sole writer of the snapshot table.
- **`tests/unit/test_config.py`** — added the missing `fetch-mlb-probables` entry to the expected `bootstrap_jobs` default (pre-existing CI failure from #377).

The issue's date-boundary concern was confirmed a non-issue: the window filter (`now <= commence_time <= end`) excludes prior-day games regardless of MLBAM's ±1-day over-fetch.

### 2. Repository pattern over the read path

Layers the path along its three real boundaries so the persistence type stops leaking outward and the live/cached read paths can't drift apart. This matches the project's Model Types convention (SQLModel = DB, Pydantic = API responses).

- **`packages/odds-core/odds_core/mlb_data_models.py`** — `select_latest_in_window`: the single latest-per-`game_pk` window-selection rule, as a pure function over domain records.
- **`packages/odds-lambda/odds_lambda/storage/mlb_pitcher_reader.py`** — `get_latest_in_window` returns `MlbProbablePitchersRecord` domain records (mapped at the storage boundary via `_to_record`) and delegates selection to `select_latest_in_window`. SQL only does the coarse `commence_time` window bound; the SQLModel type no longer escapes storage.
- **`packages/odds-mcp/odds_mcp/tools/mlb.py`** — ad-hoc response dict replaced by Pydantic `ProbablePitchersResponse` / `ProbablePitcherGame`; `fetch_status` is a validated `Literal`. Both the live and cached paths run records through the **same** `select_latest_in_window`, so results are identical *by construction*. Removes the interim `_game_dict(row: Any)` and `_latest_in_window` helpers.

This resolves the original review notes: the `Any` typing is gone (one concrete type past storage), the duplicated SQL-vs-Python "latest-in-window" logic collapses into one function, and the stringly-typed dict becomes a validated contract.

### 3. Structured MCP output schema (follow-on)

- **`packages/odds-mcp/odds_mcp/tools/mlb.py`** — the tool now returns `ProbablePitchersResponse` directly rather than `model_dump()`. FastMCP advertises the model as the tool's **structured output schema**, so agents/clients receive typed structured content instead of an opaque dict.
- **`tests/unit/test_mlb_probable_pitchers_tool.py`** — direct-call tests move to attribute access; adds MCP-level tests that drive the tool through the in-memory FastMCP client and assert (a) the advertised output schema and (b) round-tripped structured content.

## Test plan

- `uv run pytest tests/unit/test_mlb_probable_pitchers_tool.py tests/unit/test_mlb_stats_fetcher.py` — 28 passed.
- `ruff check` / `ruff format` clean; pre-commit hooks pass.

## Closes #379

### DIFF
--- a/packages/odds-core/odds_core/mlb_data_models.py
+++ b/packages/odds-core/odds_core/mlb_data_models.py
@@ -34,6 +34,29 @@ class MlbProbablePitchersRecord:
     away_pitcher_id: int | None = None
 
 
+def select_latest_in_window(
+    records: list[MlbProbablePitchersRecord],
+    start: datetime,
+    end: datetime,
+) -> list[MlbProbablePitchersRecord]:
+    """Latest record per ``game_pk`` whose ``commence_time`` is in ``[start, end]``.
+
+    The single selection rule shared by the live read-through path (records
+    straight from MLBAM) and the cached read path (records mapped from persisted
+    snapshots), so both produce identically shaped results. Records outside the
+    window are dropped; among records for the same ``game_pk`` the newest
+    ``fetched_at`` wins. Ordered by ``commence_time`` ascending.
+    """
+    latest: dict[int, MlbProbablePitchersRecord] = {}
+    for record in records:
+        if not (start <= record.commence_time <= end):
+            continue
+        existing = latest.get(record.game_pk)
+        if existing is None or record.fetched_at > existing.fetched_at:
+            latest[record.game_pk] = record
+    return sorted(latest.values(), key=lambda r: r.commence_time)
+
+
 class MlbProbablePitchers(SQLModel, table=True):
     """Snapshot of MLBAM's probable-pitcher state for a single game.
 
--- a/packages/odds-lambda/odds_lambda/jobs/fetch_mlb_probables.py
+++ b/packages/odds-lambda/odds_lambda/jobs/fetch_mlb_probables.py
@@ -1,4 +1,4 @@
-"""Fetch MLB probable-pitcher snapshots — backstop for days the agent doesn't run.
+"""Fetch MLB probable-pitcher snapshots — sole writer of ``mlb_probable_pitchers``.
 
 Flow:
 1. Hit MLB Stats API ``/schedule`` with the ``probablePitcher`` hydration for
@@ -7,10 +7,11 @@ Flow:
    (idempotent on ``(game_pk, fetched_at)``).
 3. Alerts on failure via ``job_alert_context``.
 
-The MCP ``get_probable_pitchers`` tool is the primary write path; this cron
-exists to keep the announcement-timing series populated on days the agent
-doesn't fire. The 3-day window matches MLBAM's empirical populate rate
-(D+0 fully populated, D+1..D+3 partial, D+4 onwards empty).
+This cron is the only write path for the snapshot table. The MCP
+``get_probable_pitchers`` tool is read-only — it read-throughs MLBAM live for
+agent freshness but never persists, so the agent's read-mostly DB role needs
+no write grant on the table. The 3-day window matches MLBAM's empirical
+populate rate (D+0 fully populated, D+1..D+3 partial, D+4 onwards empty).
 """
 
 from __future__ import annotations
--- a/packages/odds-lambda/odds_lambda/storage/mlb_pitcher_reader.py
+++ b/packages/odds-lambda/odds_lambda/storage/mlb_pitcher_reader.py
@@ -5,11 +5,35 @@ from __future__ import annotations
 from datetime import datetime
 
 import structlog
-from odds_core.mlb_data_models import MlbProbablePitchers
+from odds_core.mlb_data_models import (
+    MlbProbablePitchers,
+    MlbProbablePitchersRecord,
+    select_latest_in_window,
+)
 from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
 
 logger = structlog.get_logger(__name__)
+
+
+def _to_record(row: MlbProbablePitchers) -> MlbProbablePitchersRecord:
+    """Map a persisted snapshot row to the in-memory domain record.
+
+    Keeps the SQLModel persistence type from escaping the storage layer — the
+    reader speaks in domain records, the same type the fetcher produces.
+    """
+    return MlbProbablePitchersRecord(
+        game_pk=row.game_pk,
+        commence_time=row.commence_time,
+        fetched_at=row.fetched_at,
+        home_team=row.home_team,
+        away_team=row.away_team,
+        game_type=row.game_type,
+        home_pitcher_name=row.home_pitcher_name,
+        home_pitcher_id=row.home_pitcher_id,
+        away_pitcher_name=row.away_pitcher_name,
+        away_pitcher_id=row.away_pitcher_id,
+    )
 
 
 class MlbPitcherReader:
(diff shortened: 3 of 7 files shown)


Acusick1 (commented):
## Review — comment only (co-authored by Claude, can't approve)

Implements #379 faithfully. The architectural decision lands well: the agent-facing tool is now pure-read, the cron is the sole writer, and the read-mostly `odds_agent` role no longer needs (and shouldn't have) a write grant on the table. CI is green.

**Strengths**
- Control flow is clean: live path early-returns `"live"`; `httpx.HTTPError` falls through to the cached DB read as `"stale_db_only"`; `refresh=False` reads cached as `"db_only"`. The now-unused `records: list[...] = []` initializer was correctly removed — nothing references it on the fallback path.
- `_game_dict` consolidates the previously-inlined formatting so live records and DB rows are serialized through one path (good de-duplication; both `MlbProbablePitchersRecord` and `MlbProbablePitchers` share the structural interface it accesses).
- `_latest_in_window` faithfully mirrors `MlbPitcherReader.get_latest_in_window` (window filter → latest-per-`game_pk` → `commence_time` asc), so live and cached results are shaped identically.
- Test coverage is thorough: `live` (no-write), duplicate-`game_pk` collapse, `db_only`, and `stale_db_only` fallback are each exercised, plus ordering. Bundling the pre-existing `test_settings_defaults` fix (the `bootstrap_jobs` default drifted from the test in #377) is a fair scope inclusion.

**Non-blocking notes**

1. **`_game_dict(row: Any, ...)`** — `Any` discards the type safety the codebase otherwise enforces. Since it's deliberately polymorphic over two known types, a `Protocol` (or `MlbProbablePitchersRecord | MlbProbablePitchers`) would document the contract and catch field drift. Minor.

2. **Semantic shift worth being explicit about:** the old write-through persisted then DB-read, so the result was effectively a *union* of fresh records and any older cached in-window snapshots. The new live path returns **only** what the current MLBAM response contains in-window. If MLBAM ever returns a partial (non-error) response that omits an in-window game still present in cache, that game now drops from the `"live"` result rather than being back-filled from cache. I'd argue the new behavior is *more* correct for a betting agent (current truth, no silent stale-mixing), but it is a behavior change — fine to keep, just flagging it.

3. **Tie-break on duplicate `game_pk`:** with all live records stamped identical `fetched_at=now`, `_latest_in_window`'s `>` comparison means first-seen wins on a tie (arbitrary among identical-stamp dupes). This matches the DB path's arbitrariness under `DISTINCT ON`, so it's consistent — no action needed.

Nothing here blocks merge; (1) and (2) are the only things worth a second's thought.

Acusick1 (commented):
## Review — LGTM (comment-only; Claude co-authored, can't approve)

Solid, well-layered PR. The architectural framing is right: `mlb_probable_pitchers` is pipeline-ingested reference data, so making the agent-facing tool read-only and leaving the cron as sole writer is the correct fix rather than granting INSERT to `odds_agent`. CI is green on the latest commit (the first commit's failure was the pre-existing `test_config` issue, fixed in commit 2).

### What I verified
- **Control flow** (`tools/mlb.py:145-167`): `fetch_status` is assigned on every branch — `live` returns early, the `httpx.HTTPError` path sets `stale_db_only`, the non-refresh path sets `db_only`. No unbound-variable path.
- **Single selection rule**: live and cached paths both run through `select_latest_in_window`, so results are identical by construction — that genuinely resolves the prior drift concern. The live path filters the fetcher's ±1-day over-fetch back to `[now, end]`, confirming the issue's date-boundary item is a non-issue.
- **No stale callers**: `get_latest_in_window` is consumed only by the tool, so the return-type change (`MlbProbablePitchers` → `MlbProbablePitchersRecord`) has no other blast radius. The SQLModel type no longer escapes storage.
- **Writer fully dropped** from the tool; no dead imports left behind.
- Structured-output tests drive the real FastMCP client and assert both the advertised schema and round-tripped content — good behavior-level coverage.

### Minor (non-blocking)
- **Lost DISTINCT ON push-down** (`mlb_pitcher_reader.py`): the reader now `SELECT`s every in-window snapshot row and dedups per `game_pk` in Python, where it previously did `DISTINCT ON (game_pk)` in SQL. With a daily cron and a 48h default window that's ~2-3 rows/game — negligible. Worth a mental note if the cron cadence ever increases or the window widens, since transfer then scales with snapshot count rather than game count.
- **Redundant window re-filter**: SQL already bounds `commence_time`, then `select_latest_in_window` re-checks it. Harmless, and it's the deliberate price of sharing one rule across both paths — fine as-is.
- **fetched_at tie-break**: `select_latest_in_window` uses `>`, so on equal `fetched_at` the first record in input order wins. Deterministic per input; the duplicate-collapse test relies on identical records so it doesn't matter, but flagging that the ordering is input-dependent.

Nothing here blocks merge.

PR #382, opened 2026-06-01
Make deployed-vs-local job state discoverable

Acusick1:
## Why

Asked the simple question "what's deployed vs running local-only?" — and the honest answer was *you can't tell easily*. The split was spread across `eventbridge.tf`, `jobs.py`, and `config.py` with no human-readable summary, and the `DEPLOYMENT.md` EventBridge table had drifted from the Terraform (it claimed `fetch-oddsportal-results` was a fixed `cron(0 8)` rule; it's actually a self-scheduling scraper rule, gated off by default).

## What

**Single source of truth** — `docs/DEPLOYMENT.md` gains a **Job Deployment Matrix** (job → AWS/local backend, schedule type, gating var) and the stale EventBridge section is corrected.

**Orientation** — `CLAUDE.md` gains a "What runs where" block. Surfaces that the committed Terraform deploys a *narrow* AWS surface: scheduler Lambda runs only `fetch-odds-epl`, `fetch-scores-epl`, `update-status`, `check-health` (all `enable_*` gates default off; `sport_configs` is EPL-only). The scraper, agent, digest, Betfair, and **all MLB jobs** are local-only.

**Live query** — `odds scheduler list-jobs --backend aws` now inspects the deployed EventBridge schedule on demand, showing the raw `ScheduleExpression` + state. Because Terraform `ignore_changes` the dynamic rules' schedule/state, this is the only way to see the real cadences. Mechanics:
- `--backend` override on `list-jobs` (no `.env` edit needed)
- `schedule_expression` field added to `ScheduledJob` (optional, defaults `None`); a "Schedule" column shows only for the AWS backend so local listings stay clean
- `rule_prefix` added to typed `AWSConfig`, threaded through the backend factory
- env fallbacks (`RULE_PREFIX`, `LAMBDA_ARN`, `AWS_REGION`) preserved, so the Lambda runtime path is unchanged

## Testing

- `uv run pytest -k "scheduler or backend"` — 31 passed
- `ruff check` / `ruff format` clean
- `odds scheduler list-jobs --backend local` verified end-to-end (no Schedule column, as intended)
- Not exercised against the live AWS account (needs credentials); the local path and unit tests cover the wiring.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -28,6 +28,10 @@ AWS_REGION=us-east-1
 # Example: arn:aws:lambda:us-east-1:123456789012:function:odds-scheduler
 AWS_LAMBDA_ARN=
 
+# EventBridge rule name prefix (terraform `rule_prefix`, e.g. "odds").
+# Required to query the deployed schedule: `odds scheduler list-jobs --backend aws`
+AWS_RULE_PREFIX=
+
 # =============================================================================
 # MODEL LOADING (required for local score-predictions)
 # =============================================================================
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -16,14 +16,16 @@ Sport-agnostic betting data pipeline with sport-specific LLM agents. Infrastruct
 
 ### What runs where (deployed vs local)
 
-The committed Terraform deploys a **narrow AWS surface**: the scheduler Lambda runs only
-`fetch-odds-epl`, `fetch-scores-epl`, `update-status`, and `check-health` (all `enable_*`
-gates default off; `sport_configs` is EPL-only). **Everything else is local-only** — the
-OddsPortal scraper, the betting agent, daily digest, ESPN fixtures, Betfair Exchange, and
-**all MLB jobs** — driven by `odds scheduler start` (residential IP avoids Cloudflare). The
-agent is in interactive evaluation, not autonomous production. For the full job→backend
-mapping see the [Job Deployment Matrix](docs/DEPLOYMENT.md#job-deployment-matrix); for the
-live deployed schedule run `odds scheduler list-jobs --backend aws`.
+Two AWS Lambdas (`odds-scheduler`, `odds-scheduler-scraper`) plus a local scheduler
+(`odds scheduler start`). **The committed tfvars are not a reliable record of the deployed
+state** — the live account has been applied with gate/sport values that differ from the
+variable defaults, and Terraform `ignore_changes` the self-scheduling rules. The betting
+agent, daily digest, ESPN fixtures, and MLB probables are **always local-only** (no
+EventBridge rule exists for them); the agent is in interactive evaluation, not autonomous
+production. Most data-fetching jobs (odds, scores, OddsPortal scraper, Betfair) *may* run
+on AWS, locally, or both. For the design see the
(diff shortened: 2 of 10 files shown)


Acusick1 (commented):
## Review — `Make deployed-vs-local job state discoverable`

Reviewed the diff, traced the wiring, and verified the doc claims against `deployment/aws/terraform/eventbridge.tf` and the live-AWS path. CI is green on all three commits. This is a clean, well-scoped change — mostly docs (accurate) plus minimal code wiring. **(Co-authored by Claude → comment only, not approving.)**

### Verified
- **Doc accuracy holds.** `fixed_schedule_expressions = {}`, `per_sport_scheduler_jobs = ["fetch-odds", "fetch-scores"]` (+betfair when `betfair_enabled`), `global_jobs = ["update-status", "check-health"]`, scraper jobs gated by `enable_oddsportal_scraper`, `sport_configs` EPL-only. The "always local-only" set (`agent-run`, `daily-digest`, `fetch-espn-fixtures`, `fetch-mlb-probables`, inline `score-predictions`) is genuinely absent from both `scheduler_rules_map` and `scraper_rules_map`. Matrix matches.
- **Lambda runtime path preserved.** Factory now passes `settings.aws.{region,lambda_arn,rule_prefix}` (AWS_-prefixed via Pydantic), and `aws.py` keeps the `os.getenv("LAMBDA_ARN"/"RULE_PREFIX"/"AWS_REGION")` fallbacks the Lambda relies on (those env vars aren't AWS_-prefixed in the deployed function). Since `AWS_LAMBDA_ARN`/`AWS_RULE_PREFIX` are unset in Lambda, settings resolve to `None` and the unprefixed fallbacks win — behavior unchanged. Good.
- **Sort safety.** I initially suspected `sorted(key=lambda j: (j.next_run_time is None, j.next_run_time))` would `TypeError` once the AWS backend surfaced multiple rules with `None` next-run (rate-based / DISABLED-placeholder rules don't match the cron regex). Confirmed it's fine: tuple comparison short-circuits on `==` before reaching `None < None`. No issue.

### Minor / nits (non-blocking)
1. **Dead docstring in `list_jobs`** (`scheduler.py:399-411`). `app_settings = get_settings()` precedes the triple-quoted string, so it's a no-op string expression, not the function's `__doc__`. Pre-existing, but this PR edits that block (adds the Schedule line) — cheap to fix by moving the docstring above the assignment.
2. **Dual env-var naming.** `rule_prefix` now joins `lambda_arn` in having two names for one setting — `AWS_RULE_PREFIX` (settings) vs `RULE_PREFIX` (`os.getenv` in `aws.py`). Intentional and documented, and it's what preserves the Lambda path, but it's a latent footgun for anyone setting only one. Consider a one-line comment in `aws.py` noting the unprefixed names are the Lambda-runtime contract.
3. **No test for the additive surface.** `schedule_expression` population and the `show_schedule` column logic in `_print_scheduled_jobs` are pure and unit-testable without AWS creds (construct `ScheduledJob(..., schedule_expression=...)` and assert the column appears/hides). Low value given the additive default, but trivial to add if you want regression cover on the "local listings stay clean" guarantee.

Nothing here blocks merge.

PR #387, opened 2026-06-02
Unify self-scheduling into one sport-aware decision module

### LINKED ISSUE #385: Consolidate three parallel self-scheduling engines into one sport-aware decision module
## Goal

Self-scheduling jobs reimplement "proximity-to-game → polling cadence + whether-to-run" three incompatible ways, with an inconsistent backend-call convention, inconsistent overnight handling, and a latent sport-awareness bug. Unify them into a single decision engine so behavior is consistent, testable, and a clean foundation for season-aware scheduling (separate issue). No behavior change beyond fixing the sport-awareness bug.

## Background — the three engines (all map proximity → cadence)

- `SchedulingIntelligence` (`intelligence.py:37-228`) — returns a `ScheduleDecision(should_execute, reason, next_execution, tier)`; has a gate; uses `FetchTier` intervals; **not sport-aware** (`get_closest_game` queries all sports). Used by `fetch_odds`, `fetch_scores`, `update_status`.
- `ProximitySchedule` + `get_next_kickoff(sport)` (`helpers.py:16-69`) — returns an interval float; **no gate** (always runs); per-job intervals; sport-aware. Used by `fetch_oddsportal`, `fetch_betfair_exchange`.
- Hand-rolled `TIER_*` constants + `_compute_wake_interval` (`agent_run.py:34-84`) — 5 bespoke tiers with different boundaries (6/24/48h vs FetchTier's 3/12/24/72h); sport-aware; plus an `agent_wakeups` override. Used by `agent_run`.

**Additional inconsistencies to resolve:**
- Backend call: `self_schedule()` helper (`helpers.py:97-126`) used by oddsportal/betfair/agent vs direct `backend.schedule_next_execution()` used by odds/scores/status/results.
- Overnight: `apply_overnight_skip` applied by oddsportal + agent, **omitted by betfair**, and `fetch_oddsportal_results` hardcodes `tomorrow 08:00` (`fetch_oddsportal_results.py:334-335`) ignoring overnight entirely.
- Pre-schedule pattern: PRE+POST (oddsportal/betfair), POST-only (odds/scores/status), `finally` (results), PRE+override (agent).
- **Sport bug:** `SchedulingIntelligence` gates on the closest game across all sports — wrong for per-sport compound jobs (e.g. `fetch-odds-epl` is kept "in season" by MLB fixtures).

## Location

- New: `packages/odds-lambda/odds_lambda/scheduling/decision.py` (the unified engine).
- Refactor: `scheduling/intelligence.py`, `scheduling/helpers.py` (retire `ProximitySchedule`; keep/relocate `get_next_kickoff`, `apply_overnight_skip`, `self_schedule`).
- Jobs: `fetch_odds.py`, `fetch_scores.py`, `update_status.py`, `fetch_oddsportal.py`, `fetch_oddsportal_results.py`, `fetch_betfair_exchange.py`, `agent_run.py`.

## Scope

- Define one `ScheduleDecision` (keep current shape: `should_execute`, `reason`, `next_execution`, `tier`).
- Define a per-job `CadenceConfig` (intervals per canonical `FetchTier` tier) replacing `ProximitySchedule` and the agent `TIER_*` constants. All jobs share the **same tier boundaries** (`FetchTier`); only interval values differ per job.
- Provide two sport-aware decision functions sharing the query layer, overnight handling, and return type:
  - `decide_forward(sport, cadence, overnight=None, lookahead_days=...)` — keys off `get_next_kickoff(sport)`; gates `should_execute=False` when no upcoming game; returns proximity-tier `next_execution`.
  - `decide_backward(sport, window, active_interval, idle_interval)` — keys off recent games (preserve `should_execute_scores`'s 3-day and `should_execute_status_update`'s 4h semantics), made sport-aware.
- Make all proximity/recent-games queries **sport-scoped** (accept `sport_key`); fix the all-sports bug.
- Standardize every self-scheduling job on `self_schedule()` (retire direct `backend.schedule_next_execution` calls in jobs).
- Apply `apply_overnight_skip` consistently for all forward jobs that have an overnight window (add it to betfair; replace results' hardcoded 08:00 with the shared path or document why results stays fixed).
- Standardize the pre-schedule-before-work + post-reschedule pattern for the I/O-heavy jobs (oddsportal, betfair, agent already do PRE+POST; align odds/scores).
- Preserve `agent_run`'s `agent_wakeups` override behavior (it composes on top of the unified decision; only the tier math is replaced).

## Constraints

- No season/offseason logic in this issue — that is the follow-up. Behavior should be unchanged except the sport-awareness fix.
- Preserve self-scheduling-chain crash-survival (pre-schedule before I/O; reschedule even on failure where jobs do so today).
- All decision functions take an injectable `now` and `session_factory` for testing; `datetime.now(UTC)` only.
- Async throughout; no new dependencies. Backend-agnostic (local/AWS/Railway).

## Out of scope (call out, do not change)

`fetch_polymarket` (NBA-only, deprioritized, uses its own `calculate_tier` path), `check_health` (fixed 60-min global heartbeat), and the pure-cron jobs `daily_digest`/`fetch_mlb_probables`/`fetch_espn_fixtures`/`score_predictions` (no self-scheduling).

## Success criteria

- [ ] One module owns proximity→cadence and the run/skip gate; `ProximitySchedule` and the agent `TIER_*` constants are removed.
- [ ] Per-sport jobs gate on *their own* sport's next game (regression test proving EPL fetch is unaffected by MLB fixtures).
- [ ] All self-scheduling jobs call `self_schedule()`; no direct `backend.schedule_next_execution` in job modules.
- [ ] Overnight skip applied consistently (betfair included); no hardcoded wake times that ignore overnight windows except where explicitly documented.
- [ ] `agent_run` produces the same override behavior via `agent_wakeups`.
- [ ] Existing scheduling tests pass, updated where signatures change: `test_scheduler_e2e.py::test_job_self_scheduling_chain` (tier intervals), `test_local_backend.py`, `test_job_routing.py`.
- [ ] New class-based unit tests for `decide_forward`/`decide_backward` covering each tier, the no-game gate, and sport scoping (shared fixtures in `conftest.py`).

## Open questions

- Agent tier boundaries (6/24/48h) differ from canonical `FetchTier` (3/12/24/72h). Migrate the agent to canonical boundaries (slight cadence change, simpler) or let `CadenceConfig` carry a custom boundary set? Implementer to decide and flag in the PR.
- Should `fetch_polymarket` be folded into the unified engine now or left for a later pass? (Recommend later.)

## Notes

- Local backend pickles the job callable (`module.qualname`) + `JobContext` into Postgres and invokes it at fire time (`local.py` `_write_schedule`); AWS stores only the job name (`aws.py` `_schedule_with_retry`); Railway is a no-op. The unified engine must therefore live in code the jobs call, not in a dispatch wrapper.
- `run-once` (CLI, `scheduler.py:275`) invokes jobs directly, bypassing the backend — keep that path working.

Acusick1:
## Summary

Consolidates three divergent self-scheduling engines (`SchedulingIntelligence`, `ProximitySchedule`, and the agent's hand-rolled `TIER_*` constants) into a single sport-aware module, `scheduling/decision.py`.

- **`ScheduleDecision`** (`should_execute`, `reason`, `next_execution`, `tier`) and a per-job **`CadenceConfig`** keyed off the canonical `FetchTier` boundaries — all jobs share boundaries, only interval values differ.
- **`decide_forward`** — next-game proximity gate + tier (replaces `SchedulingIntelligence.should_execute_fetch`).
- **`decide_backward`** — recent-games gate (replaces `should_execute_scores` and `should_execute_status_update`).
- **`decide_forward_resilient`** — soonest-of-many-sports with a `db_fallback` cadence on query failure, so a DB outage can't break the self-scheduling chain; used by the always-run jobs (`fetch_odds`, `fetch_oddsportal`, `fetch_betfair_exchange`).
- **`OVERNIGHT_WINDOWS`** — single shared sport→window map.

**Bug fix:** all proximity / recent-games queries are now **sport-scoped**. Previously a per-sport compound job (e.g. `fetch-odds-epl`) could be kept "in season" by another sport's fixtures (regression test added).

All jobs now route through `self_schedule()` with the PRE+POST reschedule pattern (no direct `backend.schedule_next_execution` in job modules). Removed `intelligence.py`, `ProximitySchedule`, `get_closest_game`, `is_nba_season`.

## Behaviour notes
- Agent tier boundaries migrated to canonical `FetchTier` (3/12/24/72h); slight cadence shifts on a few bands. Agent is local/interactive-eval only.
- `_classify` uses strict `<` per the documented `FetchTier.max_hours` contract; this intentionally differs from `tier_utils.calculate_tier`'s `<=` at exact boundaries (pinned by a boundary-value test).
- `fetch_oddsportal_results` keeps its fixed daily 08:00 UTC wake but now self-schedules via `self_schedule`.

## Testing
- New `test_decision.py` (tier selection, overnight skip, sport scoping, backward gate, boundary values).
- Updated affected job/scheduler tests; deleted `test_scheduling_intelligence.py`.
- Full CI suite green.

## Closes #385


Acusick1 (commented):
## Review — strong refactor, a few minor follow-ups

Read against issue #385. This delivers every success criterion: one module owns proximity→cadence + the run/skip gate, `ProximitySchedule`/`intelligence.py`/agent `TIER_*` are gone, all jobs route through `self_schedule()`, the sport-scoping bug is fixed with a regression test, overnight is applied consistently (betfair included), and `decide_forward`/`decide_backward` have class-based unit tests covering tiers, the no-game gate, boundary values, and sport scoping. Verified `tests/unit/test_decision.py` passes locally (20/20) and CI is green on the head commit.

Parity check vs the deleted `SchedulingIntelligence` looks right: `fetch_odds` CADENCE (0.5/3/12/24/48, no_game=24) matches the old `tier.interval_hours` + daily check; backward windows (3d scores, 4h status) and semantics are preserved and now sport-scoped. The strict-`<` boundary divergence from `calculate_tier` is intentional, documented, and pinned by a boundary test. Nicely done.

### Minor (non-blocking)

1. **`decide_forward_resilient` log/reason names the wrong sport on combined runs.** It computes `soonest` across all `sport_keys` but then builds the decision via `decide_forward(sport_keys[0], ...)`, so the `reason` string (`f"Next {sport_key} game in ..."`) always says `sport_keys[0]` even when the soonest kickoff belongs to another sport. Only affects the no-`--sport` local multi-league runs, and it's cosmetic (logs only), but it can mislead during debugging.

2. **Backward jobs lack the DB-failure fallback that forward jobs now have.** `fetch_odds` and the always-run jobs use `decide_forward_resilient` (db_fallback cadence on query failure, chain survives). `fetch_scores`/`update_status` call `decide_backward` directly via `_scores_decision`/`_status_decision`, which run *before* the pre-schedule — so a transient DB error in the gate query throws before anything is scheduled and breaks the self-scheduling chain (relevant on AWS, where EventBridge scheduling is independent of the RDS data DB). This matches the old behavior (no regression), and #385 scoped resilience to "where jobs do so today," so it's fair to leave — but given the PR's unification theme it's worth a follow-up to give the backward gate the same crash-survival guarantee.

3. **Redundant kickoff re-query when no games exist.** In `decide_forward_resilient`, if every sport returns `None`, `soonest` is `None` and the subsequent `decide_forward(sport_keys[0], ..., next_kickoff=None)` re-runs `get_next_kickoff(sport_keys[0])` (one extra query). Harmless, easy to skip.

### Nits

- `SessionFactory = Callable[[], object]` doesn't capture the async-CM protocol it's used as (`async with factory()`); a `Protocol` or `AbstractAsyncContextManager` alias would type the contract.
- `_classify`'s `hours_until < 0 → IN_PLAY` branch is effectively unreachable through the real query path (`get_next_kickoff` filters `commence_time > now`); it only fires for injected kickoffs in tests. Fine to keep, just noting it's not exercised in production.

Not approving since the PR is co-authored by Claude — flagging the above for your call. Nothing here blocks merge.

Acusick1:
Pushed `323b46b` addressing the review (CI green):

- **Correct sport in the resilient decision.** Extracted a pure `_decide_from_kickoff` core; `decide_forward_resilient` now names the sport that actually owns the soonest kickoff (was hardcoded to `sport_keys[0]`) and no longer re-queries `get_next_kickoff` redundantly when no game exists.
- **Injectable kickoff clock.** Threaded `now` through `get_next_kickoff` so the proximity gate is fully time-controllable. This also fixes a latent flake in `test_sport_scoping_queries_own_sport`: it pinned `now=NOW (12:00 UTC)` but `get_next_kickoff` used real `datetime.now(UTC)`, so the test silently broke once wall-clock passed the fixture's hardcoded 13:00 kickoff. (The green CI run on the prior head happened to execute at 10:33 UTC, before that boundary.)
- **Tightened `SessionFactory` typing** to `AbstractAsyncContextManager[AsyncSession]`.
- **Added pure-unit coverage** for `decide_forward_resilient` — soonest-sport selection and the `db_fallback` cadence on query failure.

Left for a follow-up (out of scope here): `fetch_scores`/`update_status` backward gates still lack the DB-failure fallback the forward jobs have, and there's a pre-existing event-loop test-isolation flake (`test_self_schedules_when_process_results_raises` trips `Future attached to a different loop` when the full suite runs locally; doesn't reproduce in CI).

Acusick1:
Pushed `f806216` (CI green) — followed through on the two items from the review thread:

**1. Uniform backward-gate resilience.** Added `decide_backward_resilient`, mirroring `decide_forward_resilient`: runs if *any* sport has a recent game needing updates, and on a recent-games query failure falls back to `should_execute=True` at a `db_fallback` cadence so the self-scheduling chain survives a DB outage. `fetch_scores` and `update_status` now route through it, so the backward jobs get the same crash-survival guarantee the forward jobs already had. (Tradeoff, matching the forward jobs: on a DB failure the job fires rather than going dark — `fetch_scores` will burn a `/scores` call that then also fails, but the chain is preserved by the pre-schedule.) Added pure-unit coverage for both the any-sport gate and the fallback path.

**2. The local test flake is its own issue.** Root-caused it to a test-harness fragility — `tests/conftest.py` shares one session-scoped async engine across the whole suite while each async test runs on its own event loop, so a pooled connection can cross loops (`Future attached to a different loop`). It's latent (CI green; only surfaces in the full local run), and the fix is a test-infra change unrelated to this refactor. Filed as #388 rather than bundling an engine-scoping change here.

PR #394, opened 2026-06-28
Add deploy-with-confidence tooling for the local scheduler (smoke + deploy.sh + systemd)

### LINKED ISSUE #393: Add deploy-with-confidence tooling for the local scheduler (smoke command + deploy.sh + systemd)
## Goal

Make redeploying the always-on mini-PC scheduler safe and repeatable. Today there's no process supervisor, no pre-restart validation, and no way to confirm jobs still run after a code change — so large or stale diffs feel risky to ship. Add a smoke check, a deploy script, and a systemd unit so a redeploy either proves the changed jobs run green and then restarts, or refuses to restart and leaves the working scheduler untouched.

## Scope

- Add a `scheduler smoke` CLI command (`odds-cli`) that runs each bootstrapped job once and reports pass/fail.
- Add a `deploy.sh` at the repo root that pulls, syncs, migrates, smokes, and restarts via systemd — aborting before restart if smoke fails.
- Add an `odds-scheduler.service` systemd unit (with install/usage docs).
- Add a tag in the job registry distinguishing smoke-unsafe jobs (`agent-run`) and outward-posting jobs (`daily-digest`).

## Location

- `packages/odds-cli/odds_cli/commands/scheduler.py` — new `smoke` command (alongside `run-once`, `start`).
- `packages/odds-lambda/odds_lambda/scheduling/jobs.py` — smoke-exclusion tag(s); reuse `expected_compound_job_names()`.
- `deploy.sh` (repo root, new).
- `odds-scheduler.service` (new; document in `docs/DEPLOYMENT.md`).

## Implementation approach

**`scheduler smoke`:**
- Derive the job set from `app_settings.scheduler.bootstrap_jobs` at runtime — single source of truth shared with `start`. Expand per-sport via `expected_compound_job_names(bootstrap_jobs, configured_sports)`.
- Run each job via `get_job_function(base_name)` with a `JobContext`, wrapped per-job in try/except so one failure doesn't abort the rest. Collect results, print a pass/fail table, exit non-zero if any failed.
- Tagging: `_SMOKE_EXCLUDE = {"agent-run"}` skipped by default (never place paper bets during a deploy check); `daily-digest` tagged as outward-posting, dropped under `--no-side-effects`.
- Flags: `--include-agent`, `--no-side-effects`, `--only JOB...`, `--exclude JOB...`.

**`deploy.sh`:**
```
git pull → uv sync → uv run alembic upgrade head
→ SCHEDULER_DRY_RUN=true uv run odds scheduler smoke   # abort on failure
→ sudo systemctl restart odds-scheduler
→ systemctl status + uv run odds scheduler list-jobs
```
- `SCHEDULER_DRY_RUN=true` makes `self_schedule` a no-op (`backends/local.py` early-returns when `dry_run`), so smoke exercises real job work without writing to the live Postgres schedule store. This yields the key property: the running scheduler keeps serving old code until the new code passes smoke.

**`odds-scheduler.service`:** `ExecStart=uv run odds scheduler start`, `Restart=always`, `WorkingDirectory` = repo, `EnvironmentFile=.env`.

## Constraints

- Async throughout; jobs run via `asyncio.run` as `run-once` already does.
- No new runtime dependencies.
- Smoke must not perturb the live scheduler's schedule store (enforced via `SCHEDULER_DRY_RUN=true`).
- `bootstrap_jobs` remains the only job-set source for the local box; do not introduce a parallel list.

## Open questions

- Whether to run smoke jobs in-process sequentially vs. as subprocesses (in-process is simpler and likely sufficient; implementer's call).
- Exact table/summary formatting and the precise non-zero exit semantics (e.g., exit code = number of failed jobs vs. plain `1`).
- Whether `--only`/`--exclude` should validate names against the registry and error on unknown jobs.

## Success criteria

- [ ] `odds scheduler smoke` runs every job in `bootstrap_jobs` (per-sport expanded) except `agent-run`, prints a per-job pass/fail table, and exits non-zero if any job fails.
- [ ] `--include-agent`, `--no-side-effects`, `--only`, and `--exclude` behave as described.
- [ ] Running smoke with `SCHEDULER_DRY_RUN=true` does not add, modify, or remove any rows in the APScheduler schedule store.
- [ ] `deploy.sh` aborts before `systemctl restart` if `alembic upgrade head` or `scheduler smoke` fails, leaving the running scheduler untouched.
- [ ] `odds-scheduler.service` auto-restarts the scheduler on crash and on host reboot; `systemctl restart odds-scheduler` cleanly cycles it.
- [ ] `docs/DEPLOYMENT.md` documents the systemd install and the `deploy.sh` flow.

## Notes

- `bootstrap_jobs` is local-scheduler-only — the Lambda path (`lambda_handler.py`) is event-driven via EventBridge and never reads it, so this work has no AWS impact.
- The committed `bootstrap_jobs` default does **not** include `fetch-odds`/`fetch-scores` (those run on AWS) and comments out `agent-run`/`daily-digest`. Smoke validates whatever the box actually bootstraps, so it adapts to env overrides automatically.
- Coverage caveat to document in `--help`: smoke always exercises each job's import/config/decision/schema path, but the full fetch+ingest body only runs when a job's cadence gate (`decision.should_execute`) says execute. A `--force` gate-bypass is out of scope.
- Migration ordering: `alembic upgrade head` runs before the restart, so the still-running old scheduler briefly sees the new schema. Safe for additive migrations (the norm); note this in `deploy.sh`.
- **Dry-run isolation semantics (important for the smoke implementation):** because smoke runs with `SCHEDULER_DRY_RUN=true`, it writes nothing to the `apscheduler_schedules` store and does not count as a real execution. The real scheduler, on restart, fires purely from the persisted schedules — it neither knows smoke ran nor re-runs everything because of it (future schedules wait; an overdue one-shot catches up on startup as on any restart). The flip side: any real work smoke did (e.g. a scrape that was due) is **duplicated** when the scheduler later runs that job on its own cadence. This is safe because all jobs are idempotent upserts and the `should_execute` gate means most jobs no-op during smoke unless actually due; the only cost is occasional wasted scrape/API effort. Do **not** "fix" this by letting smoke schedule for real — that reintroduces a race with the still-running old scheduler over the shared store and forfeits the untouched-until-restart guarantee.

Acusick1:
## Summary

Makes redeploying the always-on mini-PC scheduler safe and repeatable: a redeploy now proves the changed jobs run green before it restarts, or refuses to restart and leaves the working scheduler untouched.

- **`scheduler smoke` CLI command** (`packages/odds-cli/odds_cli/commands/scheduler.py`): runs each bootstrapped job once and prints a per-job PASS/FAIL/SKIP table, exiting non-zero (exit code = number of failed jobs, clamped) if any fail. Job set is derived from `app_settings.scheduler.bootstrap_jobs` and expanded per-sport via `expected_compound_job_names()` — the same single source `start` uses, no parallel list. `agent-run` is skipped by default (never place paper bets during a deploy check); `daily-digest` is dropped under `--no-side-effects`. Flags: `--include-agent`, `--no-side-effects`, `--only`, `--exclude` (the latter two validate against the registry and error on unknown names). `--help` carries the coverage caveat (smoke always exercises the import/config/decision/schema path; the full job body only runs when the cadence gate says execute).
- **Registry tags** (`packages/odds-lambda/odds_lambda/scheduling/jobs.py`): `_SMOKE_UNSAFE_JOBS` / `_OUTWARD_POSTING_JOBS` frozensets plus `is_smoke_unsafe()` / `is_outward_posting()` helpers (accept compound or base names). Also tightened `resolve_job_name`'s return type to `tuple[str, SportKey | None]`.
- **`deploy.sh`** (repo root): `git pull --ff-only → uv sync → alembic upgrade head → SCHEDULER_DRY_RUN=true odds scheduler smoke → sudo systemctl restart odds-scheduler → status + list-jobs`. `set -euo pipefail` aborts before the restart if migrate or smoke fails, so the running scheduler keeps serving old code until the new code passes smoke.
- **`odds-scheduler.service`** systemd unit (`Restart=always`, host-reboot persistence) plus a new "Always-On Local Scheduler" section in `docs/DEPLOYMENT.md` documenting install + the deploy flow.

## Safety property

Smoke runs with `SCHEDULER_DRY_RUN=true`, which makes `self_schedule` a no-op (`backends/local.py` early-returns), so it writes nothing to the APScheduler schedule store. The live scheduler is untouched until the restart, and fires purely from persisted schedules afterward.

## Tests

Added `TestSmokeTags` (tag helpers) and `TestSmokeCommand` (12 `CliRunner` tests: default agent-run skip, `--include-agent`, `--no-side-effects` digest drop, `--only`/`--exclude` by base and compound name, unknown-token errors, and exit-code semantics). These caught a markup crash in the smoke banner that is fixed here. Targeted suites + full CI green.

## Closes #393

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+#
+# Deploy-with-confidence for the always-on local scheduler.
+#
+# Pulls the latest code, syncs dependencies, applies migrations, smoke-tests
+# the bootstrapped jobs, and only then restarts the systemd-managed scheduler.
+# If migration or smoke fails, the script aborts BEFORE the restart, leaving the
+# running scheduler untouched on the old code.
+#
+# Run from the repo root on the host running odds-scheduler.service:
+#   ./deploy.sh
+#
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+cd "$REPO_ROOT"
+
+echo "==> git pull"
+git pull --ff-only
+
+echo "==> uv sync"
+uv sync
+
+# Migration ordering: alembic runs before the restart, so the still-running old
+# scheduler briefly sees the new schema. This is safe for additive migrations
+# (the norm). If a migration is destructive, take the scheduler down first.
+echo "==> alembic upgrade head"
+uv run alembic upgrade head
+
+# Smoke runs with SCHEDULER_DRY_RUN=true so self-scheduling is a no-op: jobs do
+# their real work (gated by their cadence) but write nothing to the live
+# schedule store. The running scheduler keeps serving old code until smoke passes.
+echo "==> scheduler smoke (SCHEDULER_DRY_RUN=true)"
+SCHEDULER_DRY_RUN=true uv run odds scheduler smoke
+
+# Only reached when migrate + smoke both succeed (set -e aborts otherwise).
+echo "==> restarting odds-scheduler"
+sudo systemctl restart odds-scheduler
+
+echo "==> systemctl status"
+systemctl status odds-scheduler --no-pager || true
+
+echo "==> scheduled jobs"
+uv run odds scheduler list-jobs
+
+echo "==> deploy complete"
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -236,6 +236,64 @@ docker-compose up -d
 uv run odds status show
 ```
 
+## Always-On Local Scheduler (systemd + deploy.sh)
+
+The local scheduler can run as an always-on service on a mini-PC, supervised by
+systemd and redeployed with a single script that proves the changed jobs run
+green before it cycles the process.
+
+### systemd unit
+
+`odds-scheduler.service` (repo root) runs `odds scheduler start` under systemd
+with `Restart=always`, so the scheduler comes back after a crash or host reboot.
+Edit `User`, `WorkingDirectory`, `EnvironmentFile`, and the absolute `uv` path in
+`ExecStart` to match the host, then install it:
+
+```bash
+sudo cp odds-scheduler.service /etc/systemd/system/odds-scheduler.service
+sudo systemctl daemon-reload
+sudo systemctl enable --now odds-scheduler   # start now + on every boot
+
+systemctl status odds-scheduler              # check state
+journalctl -u odds-scheduler -f              # follow logs
+```
+
+### Smoke check
+
+`odds scheduler smoke` runs each bootstrapped job once (derived from
+`scheduler.bootstrap_jobs`, per-sport expanded — the same source `start` uses)
+and prints a pass/fail table, exiting non-zero (exit code = number of failed
+jobs) if any job fails.
+
+- `agent-run` is skipped by default (never places paper bets during a deploy
+  check); pass `--include-agent` to run it.
+- `daily-digest` posts to Discord; pass `--no-side-effects` to drop it.
+- `--only`/`--exclude` accept base or compound job names (repeatable).
+- Run with `SCHEDULER_DRY_RUN=true` so self-scheduling is a no-op and the live
+  schedule store is left untouched.
+
+Coverage caveat: smoke always exercises each job's import/config/decision/schema
+path, but the full fetch+ingest body only runs when the job's cadence gate
+(`decision.should_execute`) says execute.
+
+### deploy.sh
+
+`deploy.sh` (repo root) is the redeploy entry point:
+
+```bash
+./deploy.sh
+```
+
+It runs: `git pull` → `uv sync` → `alembic upgrade head` →
+`SCHEDULER_DRY_RUN=true odds scheduler smoke` → `sudo systemctl restart
+odds-scheduler` → `systemctl status` + `odds scheduler list-jobs`. If migration
+or smoke fails it aborts **before** the restart, so the running scheduler keeps
+serving the old code untouched until the new code passes smoke.
+
+Migration ordering: `alembic upgrade head` runs before the restart, so the
+still-running old scheduler briefly sees the new schema. Safe for additive
+migrations (the norm); for a destructive migration, stop the service first.
+
 ## Expected Costs
 
 ### The Odds API
--- a/odds-scheduler.service
+++ b/odds-scheduler.service
@@ -1,0 +1,19 @@
+[Unit]
+Description=Odds pipeline local scheduler (APScheduler)
+After=network-online.target postgresql.service
+Wants=network-online.target
+
+[Service]
+Type=simple
+# Replace with the deploying user and the absolute repo path on the host.
+User=andrew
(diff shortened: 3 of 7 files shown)


Acusick1 (commented):
## Review — PR #394 (deploy-with-confidence tooling)

Solid, well-scoped work. The smoke command reuses the single `bootstrap_jobs` → `expected_compound_job_names()` source rather than a parallel list (good), mirrors `run-once`'s `JobContext`/`get_job_function` pattern, the dry-run isolation property holds, and the tag helpers/exit-code semantics are well tested. CI is green. One thing worth deciding before merge, plus a few minor notes.

*(Co-authored by Claude — leaving as a comment, not an approval.)*

### Worth deciding: `deploy.sh` will post a daily digest to Discord on every deploy

Most job `main()`s self-gate on `decision.should_execute` (e.g. `fetch_scores.py:98` returns early when not due) — that's the documented coverage caveat, and it means smoke is a cheap no-op for them unless actually due. But `daily_digest.main()` (`daily_digest.py:344`) has **no cadence gate** — it calls `send_digest()` unconditionally, which posts to Discord.

`deploy.sh` runs `smoke` **without** `--no-side-effects`. So if the box's `bootstrap_jobs` includes `daily-digest` (the `.env.example` example list does, and the digest is local-only per CLAUDE.md), every `./deploy.sh` posts a full betting digest to the shared channel at deploy time — regardless of the 08:00 cron. That's exactly the noise the `_OUTWARD_POSTING_JOBS` tag + `--no-side-effects` flag were built to prevent, but the primary consumer doesn't use them.

Suggest `deploy.sh` run `smoke --no-side-effects` (a pre-restart validation should be silent on shared channels), or confirm a deploy-time digest is intended.

### Minor

- **`--only`/`--exclude` doesn't override the default skip.** `smoke --only agent-run` selects it, then classification skips it as smoke-unsafe → nothing runs, and with an empty `to_run` the command prints "0 passed … ✓ Smoke check passed" and exits 0. An explicit `--only` arguably signals intent to run; at minimum an all-skipped run reporting success is misleading (doesn't affect `deploy.sh`, which passes no `--only`). Consider treating explicit selection as opt-in, or exiting non-zero / warning when `to_run` is empty.
- **systemd `After=...postgresql.service`** is ineffective if Postgres runs via Docker on the host (local dev uses docker-compose). `Restart=always` + `RestartSec=5` covers the startup race, so this is cosmetic, but the dependency as written won't order against a containerized DB. Worth a comment in the unit.
- **`deploy.sh` self-pull:** `git pull --ff-only` can rewrite `deploy.sh` mid-execution; bash reads scripts incrementally, so a changed script can misbehave on that run. Common gotcha, low severity — `exec`-ing a fixed bootstrap step or noting it is enough.

### Confirmed good
- `agent-run` correctly resolves through `get_job_function` (`main` exists in `_JOB_MODULE_MAP`), so `--include-agent` works.
- Test double patches `odds_lambda.scheduling.jobs.get_job_function` and relies on the in-function import binding at call time — correct, and CI proves it.
- Exit-code clamp `min(len(failed), 255)` and per-sport failure counting tested.

